PR #13, opened 2026-03-01
feat: Windows64 local multiplayer support

deadman96385:
- Skip QuadrantSignin (profile selector) on Windows64 in both LoadMenu and CreateWorldMenu, proceeding directly to local play since Xbox Live stubs always return true for IsSignedInLive()
- Fix IsLocalMultiplayerAvailable() to not require IsHiDef() on Windows64
- Allow pad-connected players to join without a profile sign-in check
- Fix ghost RemotePlayer creation by scanning all local player slots and matching on server-assigned player index rather than controller slot, fixing P3/P4 ghost entities when joining out of controller order
- Give each player a unique name (Player 1-4) based on controller index instead of a single shared stub name
- Use raw XInput (XInputGetState) for secondary controller join detection, bypassing the 4J toggle system which consumes all button presses before game logic runs; uses a 120-frame latch for a reliable detection window
- Add .gitignore for Visual Studio build artifacts and output directories


NEVARLeVrai:
oh nice ! thx
https://github.com/smartcmd/MinecraftConsoles/issues/15

void2012:
@deadman96385 there is branch conflict now as @rtm516 already added .gitignore

PR #24, opened 2026-03-01
Strip console targets

omrih4:
COMPLETELY strips non-windows x64 targets from the project
aditionally changed default startup project to Minecraft.Client


omrih4:
as far as i know VS 2022 also works with v145, this pr just targets the latest platform

omrih4:
done

Back-Slash-N:
where will the vote be held?

PR #25, opened 2026-03-01
Fix for exe not running, not founding the project directory

NEVARLeVrai:
In _tWinMain (Windows64_Minecraft.cpp) add logic to detect if the executable path contains "\\x64\\". If found, truncate the path at that position, append "\\Minecraft.Client" and call SetCurrentDirectoryA to set the process working directory. This ensures relative resource paths resolve correctly when running from an x64 build output directory; the change is guarded by a substring check and uses MAX_PATH-safe APIs.


void2012:
Please always zero-initialize arrays. Also, consider using the wide version of these functions because they avoid ANSI <-> UTF-16 conversion, to avoid needless performance hit.

PR #28, opened 2026-03-01
Fix game exit

rtm516:
Makes the game exit after confirmation instead of crashing/not doing anything

image


rtm516:
Ready to go

PR #34, opened 2026-03-01
Copy release assets instead of changing working directory at runtime

galenguyer:
Changing the working directory at runtime is hacky and breaks if you copy the Release folder elsewhere, like if you're sending a zip of it to a friend. This copies assets into the Release folder instead.


omrih4:
was this tested? it's bugged (folders inside Common are empty)

PR #37, opened 2026-03-01
Unlock DLC, fix world names in load menu, fix dlc pack titles

simulpingus:


Lambdagon:
Could use a better name and description?

void2012:
Yes I know, this would just make reviewing faster. We are going to add a PR template anyway to standardize things. Sorry for pickiness.

PR #45, opened 2026-03-01
Remove need for Durango folder, properly use Windows64Media, and add all DLC content

simulpingus:
# Pull Request Template

## Description
Remove the need for the Durango folder
For Win64, Durango is hardcoded. We can use Windows64Media instead, which is proper, because Durango is the codename for Xbox One.
Also, this copies over all DLC content from Windows and Durango to Windows64Media, allowing it all to be used.


void2012:
@simulpingus people say this causes problems, should I revert this?

PR #59, opened 2026-03-01
Add missing DLC and Sounds to Windows64Media

codeHusky:
## Description
Adds missing DLC, sounds, media from Durango to Windows64

### Previous Behavior
Game fails to launch without throwing Audio exceptions, fails to load worlds

PR #63, opened 2026-03-01
Revert "Add missing DLC and Sounds to Windows64Media"

void2012:
Reverts smartcmd/MinecraftConsoles#59


Khaminations:
why

OneAvargeCoder193:
why was this reverted? it fails to build randomly without it, and also there is no music now

PR #71, opened 2026-03-01
feat: add DInput/WinMM controller support for Windows

deadman96385:
# Pull Request

## Description
Adds support for non-XInput controllers (PS4/DS4, generic HID gamepads) on Windows by bridging the WinMM joystick API into the game's XInput-based input pipeline.

### Previous Behavior
Only XInput-compatible controllers (Xbox controllers and XInput-mode adapters) worked on the Windows build. Connecting a PS4 DualShock 4 or any other DirectInput/HID gamepad produced no input in the game.

### Root Cause
The game's input library (4J_Input.lib) is a precompiled static library that exclusively uses the XInput API (XInputGetState, XInputGetCapabilities). XInput is Xbox-specific and is completely blind to non-XInput devices — the DS4 and other
  HID gamepads register with Windows as DirectInput/WinMM joysticks and are never enumerated by XInput, so the game had no path to receive their input.

### New Behavior
Non-XInput controllers are automatically detected and assigned to empty controller slots. The DS4 and other HID gamepads work with correct button and axis mapping: face buttons, D-pad, shoulder buttons, analog sticks, and analog triggers all
  behave as they would on an Xbox controller. The device list is refreshed every ~3 seconds to handle hot-plugging.

### Fix Implementation
 Since 4J_Input.lib is statically linked, it shares the host executable's Import Address Table (IAT). At startup, DInputManager_Init() walks the PE header of the process (IMAGE_DOS_HEADER → IMAGE_NT_HEADERS → IMAGE_IMPORT_DESCRIPTOR) to find the XInput DLL import entries and overwrites the XInputGetState and XInputGetCapabilities function pointers with hooks using VirtualProtect. The hooked XInputGetState first tries the real XInput call; if a slot reports ERROR_DEVICE_NOT_CONNECTED, it checks whether a WinMM joystick has been assigned to that slot and reads it via joyGetPosEx. The WinMM state is then translated into XINPUT_STATE format — DS4 buttons are mapped via the dwButtons bitmask, the POV hat is converted to XInput DPAD flags, stick axes are scaled from WinMM's 0–65535 range to XInput's −32768–32767 (with Y axes inverted), and trigger axes (DS4 HID axis order: Z=RSX, Rz=RSY, Ry=R2, Rx=L2) are scaled to 0–255. DInputManager_Tick() is called once per frame and periodically re-runs slot assignment to pick up newly connected devices.

### DIFF
--- a/Minecraft.Client/Minecraft.Client.vcxproj
+++ b/Minecraft.Client/Minecraft.Client.vcxproj
@@ -16302,6 +16302,7 @@ xcopy /q /y /i /s /e $(ProjectDir)Durango\CU  $(LayoutDir)Image\Loose\CU</Comman
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Release|ORBIS'">true</ExcludedFromBuild>
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='ContentPackage_Vita|ORBIS'">true</ExcludedFromBuild>
     </ClInclude>
+    <ClInclude Include="Windows64\Windows64_DInputManager.h" />
     <ClInclude Include="Windows64\KeyboardMouseInput.h">
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='ContentPackage|Durango'">true</ExcludedFromBuild>
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='CONTENTPACKAGE_SYMBOLS|Durango'">true</ExcludedFromBuild>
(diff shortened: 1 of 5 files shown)


deadman96385:
Works fully on DS4 and my BT Xbox controller but haven't tested on generic ones.

Legeend28 (approved):
yea pretty good bro 10/10 legeend28 approved

WhyCron:
I'm giving it a basic attempt, I currently have SDL3 initialising and enumerating available controllers, gonna work on the part that reads inputs in a bit.

I do have XInput turned off for SDL3, so the game will still use XInput directly for those controllers and SDL3 will ignore them.

PR #78, opened 2026-03-01
Optimize file paths in some include files to use relative paths

WushR00M:
# Pull Request

## Description
Optimizes the file paths used in the includes in some files in the Common directory to use proper relative file paths instead of absolute file paths and prevent things like `..\..\Minecraft.Client\Common` when the file is *already in* Common. A small request, but I'm curious to see if we can apply this across the entire repository, would take a lot of work, but I think it can speed up compilation, can it not?

_(example image):_

image

### Previous Behavior
The code worked fine, likely took longer for the compiler to build it as it needed to calculate which directory to go to

### Root Cause
Purely design flaw, I have no idea what 4J were thinking making this, but if it worked for them, it worked for them.

### New Behavior
The code works fine, compiler *could* take shorter to build (however PCs are so fast, you might not notice a difference)

### Fix Implementation
Like the description says:
> Optimizes the file paths used in the includes in some files in the Common directory to use proper relative file paths instead of absolute file paths and prevent things like `..\..\Minecraft.Client\Common` when the file is *already in* Common.

### DIFF
--- a/Minecraft.Client/Common/Consoles_App.h
+++ b/Minecraft.Client/Common/Consoles_App.h
@@ -5,11 +5,11 @@ using namespace std;
 #include "Audio/Consoles_SoundEngine.h"
 
 #include <xuiapp.h>
-#include "..\Common\Tutorial\TutorialEnum.h"
+#include ".\Tutorial\TutorialEnum.h"
 
 #ifdef _XBOX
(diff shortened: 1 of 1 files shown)


smartcmd:
ready for merge?

PR #82, opened 2026-03-01
now can skip the tutorial

### LINKED ISSUE #29: Skip the tutorial by pressing B on Keyboard
Image

Currently we are stuck at tutorial since we can't press B, I tried pressing escape to go back but nothing works I had to use a xbox controller.

gabriel-fondato:
- Fixes #29

### DIFF
--- a/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
+++ b/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
@@ -1220,6 +1220,17 @@ int APIENTRY _tWinMain(_In_ HINSTANCE hInstance,
 			ToggleFullscreen();
 		}
 
+		// TAB opens host options menu. - Vvis :3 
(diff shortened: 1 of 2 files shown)


Minecraft.Client/Common/Tutorial/ChoiceTask.cpp
```diff
@@ -7,6 +7,7 @@
 #include "TutorialConstraints.h"
 #include "ChoiceTask.h"
 #include "..\..\..\Minecraft.World\Material.h"
+#include "..\..\Windows64\KeyboardMouseInput.h"
```
LynithDev: This should be included conditionally or preferably not at all. You're including platform dependant code into the common (shared across platforms) only code. It'll produce compile errors for other platforms.

FemOrWhat:
good pull request.

PR #62, opened 2026-03-01
Add nightly release github action

omrih4:
# Pull Request

## Description
Adds nightly release action

### Implementation
Builds the game on push to `main`, zips build and updates nightly release with the zip

Create a nightly release and tag before merging.
image

## Related Issues
- Fixes #2

PR #84, opened 2026-03-01
Create Manual Workflow for LCE Artifact Creation

Firebladedoge229:
# Pull Request

## Description
this pr adds a github workflow to create and upload both debug and release releases as github artifacts on demand, removing the need to build it locally + making it easier for everyone involved to just have the compiled binary when needed instead of having to build it themselves

### Previous Behavior
before, (as of now), there was only a workflow for nightly releases, which just built the release configuration
this action builds both branches, just incase testing between debug and release builds is needed.

### New Behavior
manual building (along with the nightly action), with both debug + release support and increased stability (windows-2022 build, first-party individual actions, etc)

### Fix Implementation
creation of a github action

## Related Issues
- Related to #62

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,35 @@
+name: Nightly Release
+
+on:
+  workflow_dispatch: 
+  push:
+    branches:
+    - 'main'
+
+jobs:
+  build:
+    name: Build Windows64
+    runs-on: windows-latest
+
+    steps:
+    - name: Checkout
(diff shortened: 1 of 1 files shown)


Firebladedoge229:
> Thanks. Yes, the `Assertion failure` error is reproducable on native Windows 10, too. Needs further investigation

aight
got the debug build working on linux, although with my current setup there is no sound coming from the game (not sure what that problem is being caused by, probably something to do with my system)

with this knowledge the action build should work fine on windows
image

Issue #53, opened 2026-03-01
[Bug]: No sound


RatCatcherVr-dev:
### Description

i got no sound when playing.

### Steps to Reproduce

1. open the game
2. try to hear sound
3. if you heard no sound well, you're in the same boat as me...

### Expected vs. Actual Behavior

no sound...

### Build Configuration

Debug

### Target Platform

Windows64

KdotDevelopment:
I ran into the same issue, it was because /x86/Release/Windows64Media/Sound wasn't present. Once I copied that over from /Minecraft.Client/Durango/Sound it started working again. This appears to be from https://github.com/smartcmd/MinecraftConsoles/pull/45

NEVARLeVrai:
> I ran into the same issue, it was because /x86/Release/Windows64Media/Sound wasn't present. Once I copied that over from /Minecraft.Client/Durango/Sound it started working again. This appears to be from [#45](https://github.com/smartcmd/MinecraftConsoles/pull/45)

just tried i have still no sound

NEVARLeVrai:
> pull the git again and rebuild

I pulled the latest changes from Git and rebuilt the project.

However, when playing in Thought Visual studio debug or release, it now throws an error on every build.

(See screenshot attached.)

The only version that works is the Release build.
However, I need to be able to execute the .exe directly from its folder location, not run it through Visual Studio.

<img width="512" height="285" alt="Image" src="https://github.com/user-attachments/assets/babec063-a7a6-4323-ad7d-15b33272f29c" />

universeindex:
this broke on wine as well (debug build), was working before
[backtrace.txt](https://github.com/user-attachments/files/25665582/backtrace.txt)
<img width="1473" height="795" alt="Image" src="https://github.com/user-attachments/assets/0bd90482-94ca-491d-aef1-fe586a2e3cfb" />

NEVARLeVrai:
[...]

PR #93, opened 2026-03-02
Fix sound not working

SleepyLark:
# Pull Request

## Description
Copies the `Sound` folder from `Durango` to `Windows64` and changed path reference in order for it to have audio playback work

### Previous Behavior
When compiled, the finished build wouldn't play any audio. Users had to copy the `Sound` folder from `Minecraft.client/Durango` to `Windows64Media` in order for it to work.

### Root Cause
[Commit 9691561](https://github.com/smartcmd/MinecraftConsoles/commit/9691561693c3171262a9ec9a0e4e8663949c3fa0) changed where the sound engine looks for the sound folder but didn't actually commit the change that copies the folder to the new path.

### New Behavior
The code looks instead to the folder `Windows64` since it matches the folder structure of `Durango` more than `Windows64Media` does (which looks to be only for DLC content). 

### Fix Implementation
- Copy `Sound` folder from `Durango` to `Windows64`
- Change SoundPath to  `char SoundEngine::m_szSoundPath[]={"Windows64\\Sound\\"};`

Alternatively, reverting the original commit works as well. Depends on if you want clarity that this is for building the PC version or not.

## Related Issues
- Related to #53

### DIFF
--- a/Minecraft.Client/Common/Audio/SoundEngine.cpp
+++ b/Minecraft.Client/Common/Audio/SoundEngine.cpp
@@ -56,7 +56,7 @@ void SoundEngine::playMusicTick() {};
 #else
 
 #ifdef _WINDOWS64
-char SoundEngine::m_szSoundPath[]={"Windows64Media\\Sound\\"};
+char SoundEngine::m_szSoundPath[]={"Windows64\\Sound\\"};
 char SoundEngine::m_szMusicPath[]={"music\\"};
 char SoundEngine::m_szRedistName[]={"redist64"};
 #elif defined _DURANGO


simulpingus:
@void2012 revert

SleepyLark:
> hi what crack are you smoking image image

Apologies, but at the time when I pulled the repo I could not find that build file you're referring to that copies the folder after building and my DurangoMedia folder didn't have a sound folder. I mainly wanted to describe the issue and offer a solution since I saw myself others were having the same problem (and if I simply posted about it on Discord no one would be able to find it).

I was also curious why it was looking in Windows64Media when none of the other paths was looking in their _Media folder for their sound in that same code block.

PR #94, opened 2026-03-02
Bound F3 and F4 to open debug menus

jeanmajid:
# Pull Request
## Description
Binds F3 to open debug info and F4 to open a Debug screen

I couldn't find any keyboard keybindings for those features, but maybe I missed it. If I missed it, then please let me know. If I didn't then this commit adds those

# Changes
## Previous Behavior
Not being able to open the screens (probably)

## Root Cause
No code

## New Behavior
When pressing F3 it shows on screen debug info and F4 opens a debug screen, where you can do some things


void2012:
Have you tested it?

PR #95, opened 2026-03-02
Create MAP.md

wize-logic:
# Pull Request
## Description
Allows you to view a repository map which helps you easily find things in the codebase.

## Changes
Added MAP.md


wize-logic:
> IMO, we should use something like Doxygen instead.

If anyone ever has anything better, the map can be updated or replaced.

PR #96, opened 2026-03-02
Implement basic multiplayer functionality

simulpingus:
# Pull Request

## Description
Implements basic LAN functionality
launch_arguments:

name: -name usage: "-name " description: Sets your in-game username.
name: -ip usage: "-ip " description: > Manually connect to an IP if LAN advertising does not work or if the server cannot be discovered automatically.
name: -port usage: "-port " description: > Override the default port if it was changed in the source.
example: command: "Minecraft.Client.exe -name Steve -ip 192.168.0.25 -port 25565"

## Changes
See commit

Directly sourced from https://github.com/LCEMP/LCEMP
image

### DIFF
--- a/Minecraft.Client/Minecraft.Client.vcxproj
+++ b/Minecraft.Client/Minecraft.Client.vcxproj
@@ -1441,22 +1441,14 @@ xcopy /q /y /i /s /e $(ProjectDir)Durango\CU  $(LayoutDir)Image\Loose\CU</Comman
     </Link>
     <PostBuildEvent>
       <Message>Copying game assets to output directory</Message>
-      <Command>mkdir "$(OutDir)music" 2&gt;nul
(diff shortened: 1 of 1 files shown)


Minecraft.Client/Common/Network/PlatformNetworkManagerStub.cpp
```diff
@@ -503,10 +670,60 @@ wstring CPlatformNetworkManagerStub::GatherRTTStats()
 
 void CPlatformNetworkManagerStub::TickSearch()
 {
+#ifdef _WINDOWS64
+	if (m_SessionsUpdatedCallback == NULL)
+		return;
+
+	static DWORD lastSearchTime = 0;
```
JPeisach: maybe we can just use "int" here because weren't in Microsoft specific code (at least I don't think)

PR #145, opened 2026-03-02
Change flight activation to be similar to java

ZincTheMetal:
# Pull Request

## Description
Removes the need to let go of the jump button to activate flight

Lowers the flight activation window from 10 ticks to 7 ticks

### Previous Behavior
You had to jump twice and let go of the jump button in the span of 10 ticks to activate flight

### New Behavior
You jump twice in the span of 7 ticks to activate flight

### Fix Implementation
Remove the !input->jumping check to activate flight and lower the jumpTriggerTime from 10 to 7


ZincTheMetal:
My bad if I've done something wrong, this is my first time contributing to a public repo

PR #148, opened 2026-03-02
Allow for loading of WAVE Audio files

GuglioIsStupid:
# Pull Request

## Description
This PR just implements a simple check for .wav files, loading it if the .binka file is not found, making it easier for audio replacements

Simply just a QOL feature, I like the idea of being able to easily replace audio without having to jump through a hoop or two

### Previous Behavior
Previously, the game would only check for .binka files

### New Behavior
The game now checks for both .binka and .wav files via an array, making it potentially easier for other audio formats to be added (Would need their own decoders implemented)

https://github.com/user-attachments/assets/a64bfaf9-26cc-4631-b979-27ad92932934


GuglioIsStupid:
Oh, my bad, ill updated the pr to change it!

PR #207, opened 2026-03-02
Copy assets fix

GuglioIsStupid:
# Pull Request

## Description
This fixes the CopyAssets.cmake script to properly copy over the correct files instead of copying every file

Also changed to to only copy over Minecraft.Client/Common/Media/MediaWindows64.arc instead of all Minecraft.Client/Common/Media/ files and the other platform .arc files

### Previous Behavior
Would copy everything over, making file sizes larger

### New Behavior
Only copies the correct and used files


GuglioIsStupid:
Actually, im gonna draft this real quick as I realized the same thing can be done for the language files

GuglioIsStupid:
(idk how to draft)

PR #228, opened 2026-03-03
Cleaner implementation of support dynamic resizing aspect ratio

### LINKED ISSUE #222: Latest Patch no longer works on Proton
The previous commits have worked beautifully on my Linux desktop with proton prior to the latest patch which now only nets a black screen when launching Proton. At least the sound works

I know it's probably out of the realm of this project to support Linux but if something broke in Proton, something may have also broken in windows as well.

fayaz12g:
## Description
This re-adds the dynamic aspect ratio scaling of the 3d environment with a simpler approach.

I've opened this as a draft so testing can be done on wine/Linux environments to ensure compatibility first.

Video showing it working in CrossOver on Mac: 
https://imgur.com/a/6XHePYi

### Previous Behavior
Stretching a window for ultrawide resolutions causes the game the look stretched out. 

### New Behavior
This PR fixes the window squishing/stretching by making the 3d rendering look correct based on the current window size.

### Root Cause
The previous issue seemed to do with memory management. I was creating a local value to map the window size to, then releasing it, but it caused some issues forWine/Linux users.

### Fix Implementation
Now the implementation is quite simple. The window values are now directly referenced in the WM_SIZE case, and set to the window size, then the aspect ratio is calculated in glWrapper.

Additionally, instead of using new unsigned ints, the original window size ints are used, which also means that this value does not need to be set on init.

## Related Issues
- Fixes #222

### DIFF
--- a/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
+++ b/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
@@ -431,14 +431,6 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 			return TRUE;
 		}
 		return DefWindowProc(hWnd, message, wParam, lParam);
-	case WM_SIZE:
-	{
-		// Set the screen width and height used for calculating the aspect ratio
-		g_iScreenWidth = LOWORD(lParam);
-		g_iScreenHeight = HIWORD(lParam);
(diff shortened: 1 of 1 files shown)


fayaz12g:
@void2012 Let me know if I should mark it as ready instead of a draft. I've now tested it on my mac in CrossOver and confirmed it works. @enderdude77 indicated in the issue that it is working on his end as well in Linux

lspepinho:
I might test this tomorrow too with Wine

fayaz12g:
> Great work!

Thanks! Sometimes the solution is to make it simpler rather than over engineered 😅

PR #246, opened 2026-03-03
chunk optimization

detectiveren:
## Description
Makes chunks load a bit faster, it should be treated as a temp fix for now.

## Changes
Increased the amount of chunk rebuilding around the player up to 6 chunks.

### Previous Behavior
Chunks would load at a slow pace that in creative mode even the player flying could get ahead of chunk loading

### Root Cause
The values for specific variables that influence chunk rendering were low for modern hardware standards.

### New Behavior
Chunks should load a bit more faster around the player and the player shouldn't be able to get ahead of chunk rendering.

### Fix Implementation
Increased the values for dirty chunk check periods, concurrent chunk rebuilds, chunks that can render within distance from the player and command buffer allocations.

Credit to @twest0001 as they discovered that increasing MAX_COMMANDBUFFER_ALLOCATIONS from 55 * 1024 * 1024 to 2047 * 1024 * 1024 makes chunks load a bit faster too.  That is included in this PR along with my changes.

## Related Issues
- Related to #54

### DIFF
--- a/Minecraft.Client/LevelRenderer.cpp
+++ b/Minecraft.Client/LevelRenderer.cpp
@@ -1964,8 +1964,8 @@ bool LevelRenderer::updateDirtyChunks()
 						{
 							if( (!onlyRebuild) ||
 								globalChunkFlags[ pClipChunk->globalIdx ] & CHUNK_FLAG_COMPILED ||
-								( distSq < 20 * 20 ) )	// Always rebuild really near things or else building (say) at tower up into empty blocks when we are low on memory will not create render data
-							{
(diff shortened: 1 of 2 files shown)


codeHusky:
Please pull in main, we've adjusted the Command Allocation thing in a recent commit so your change there is partially redundant. For better compat we should prob be using your way of doing it and revert Xbox One to 512M rather than the higher value it was set to by an earlier PR

detectiveren:
> If you fully fix up what's in main atm and this tests stable this should be fine to use

Ah, I hopped off my PC for today just as a finished this PR, my bad lol, would have fixed it instantly

PR #251, opened 2026-03-03
Miniaudio Implementation

GuglioIsStupid:
# Pull Request

## Description
Revamps the sound engine to use Miniaudio on Desktop, keeps .binka support but unused for Windows64 as I currently don't know if .binka would want to be kept.

### Previous Behavior
Previously only used a propriety library

### New Behavior
Now uses Miniaudio, an open sourced single-header library that is cross-platform

This commit is still WIP! It is not ready to be merged, I will resolve conflicts as time goes by as I have to redo quite a few things regarding audio


fayaz12g:
Okay, thanks! Looks like when you had a merge conflict in commit `8e5ed49`, you set everything to accept incoming change, when instead you should have accepted the current change for areas you did not touch

PR #257, opened 2026-03-03
Fix player save data issue & multiple username implementations

### LINKED ISSUE #218: Player data no longer loading
In this video, you can observe me pick up some seeds from a chest, then save and close the world. Upon reopening, I go back to world spawn with a default inventory and the item(s) are gone.

https://files.catbox.moe/xqhrh9.mp4

simulpingus:
# Pull Request

## Description
Removed the conflicting username systems, which in turn also fixes the players save data issue

### Previous Behavior
There were three implementations of usernames which created conflicts with save data

### New Behavior
One username system (username.txt)

## Related Issues
- Fixes #218

https://github.com/user-attachments/assets/a07529c5-ea05-47e9-a4d8-f43772f30eb5

### DIFF
--- a/Minecraft.Client/Extrax64Stubs.cpp
+++ b/Minecraft.Client/Extrax64Stubs.cpp
@@ -21,7 +21,6 @@
 #include "Windows64\Social\SocialManager.h"
 #include "Windows64\Sentient\DynamicConfigurations.h"
 #include "Windows64\Network\WinsockNetLayer.h"
-#include "User.h"
 #elif defined __PSVITA__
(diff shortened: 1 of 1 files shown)


simulpingus:
also, i changed the type to const char* but this does not cause any problems with existing and is probably better because usernames shouldn't be overwritten during gameplay anyways

void2012:
Maybe invite anyone else to continue working on this PR?

iLikeGothMommys:
Is this in the latest nightly? As saving still doesn't work for me.

PR #259, opened 2026-03-03
Miniaudio implementation

GuglioIsStupid:
# Pull Request

## Description
Reworks the sound engine to implement Miniaudio, a single header audio library

### Previous Behavior
Previously used Miles (Old!!! Weird to work with!!)

### New Behavior
Sound effects & Music are now handled with Miniaudio, making it easier to add more formats down-the-line

## Related Issues
- Related to #123 

Some things are still a bit wonky like the 3D space stuff, need to put more time into it..

Minecraft.msscmp file is now unused as the assets are not compressed into a file, making it easier for audio replacements. .binka files are no longer supported on Windows64 platforms unless I can figure out the format and finish BinkDecoder

Works on both Release and Debug (CMAKE! I haven't built via the vsxproject as I don't use that much)


GuglioIsStupid:
Removed Miles from loading in the source, Miniaudio is used for all platforms in it now

void2012:
Good. By the way, have you tried measuring the performance? How is it different from AIL in terms of efficiency?

GuglioIsStupid:
Finally figured out the issue! I was doing some bullshit in the spatial position code :P 

Audio looks delayed in the video, but its just my OBS. It delayed audio for some reason

https://github.com/user-attachments/assets/fbd98776-3b25-44b4-846c-66c63fcbf12d

Builds on Debug and Release via CMake, unknown if sln does as idk if i need to manually include miniaudio.h there.

void2012:
Try compiling with .sln, if it doesn't find the miniaudio headers, then add them in `Additional includes`

codeHusky:
Files changed shows 0 ?

GuglioIsStupid:
Yeah I got genuinely no idea why it does that LOL I only edited SoundEngine.cpp/hpp and added miniaudio.h

codeHusky:
Could you make a new PR or otherwise fix this so the code changes show up in files changed? not confident in merging anything if its busted like that

GuglioIsStupid:
I'll make a new PR and see if it fixes it

PR #327, opened 2026-03-03
fix: properly offset the mouse position in containers

4wincode:
## Description
This offsets the mouse position in containers so it's at the correct location on the screen. Video: https://www.youtube.com/watch?v=cBJgRcisq60

### Previous Behavior
The mouse could not move past the left or top edges of containers without going outside the window.

### Root Cause
The mouse position was simply multiplied by the movie/client ratio with no offset.

### New Behavior
The mouse is now at the correct location anywhere on the screen, including past the left or top edges of containers. It's also centered properly when opening containers.

### Fix Implementation
The X and Y position of `m_controlMainPanel` is subtracted from the mouse position. There might be a better way of doing this, but this seems to work just fine.


4wincode:
I completely forgot to link that again! I've updated the description to include it.

PR #333, opened 2026-03-03
Removed some private information

TygodeVries:
## Description
Was reading some documents, contained some phone number, name and email, decided it might look better if we removed it.

## Changes
Used word to remove a phone number and email from the Ratings_Submission_Form.doc


JPeisach:
It will still be in the git tree though, I'm not sure how git magic can be done to make it look like the file never existed but it would most likely require force pushing

PR #335, opened 2026-03-03
Fix Texture Pack images in menu

Zekken01:
## Description
Add _WINDOWS64 to the HD platform check for substitution texture sizes so icons display correctly.

### Previous Behavior
Screenshot_4

### Root Cause
_WINDOWS64 was missing from the #if that sets 96×96 for HD platforms. The HD SWF skins on Windows64 use 96×96 icon placeholders, so the 64×64 size caused mismatched scaling.

### New Behavior
7yaKDyB

### Fix Implementation
Added `|| defined _WINDOWS64` to the existing `#if (defined __ORBIS__ || defined _DURANGO) `


rtm516:
Also fixes the DLC world icons

PR #340, opened 2026-03-04
Fix: Sorted the item list in the debug overlay

CS-DaviMagalhaes:
## Description
This change sorts the Item list alphabetically in the debug overlay (f4). Now finding specific blocks or items is faster. Before you would need to go down every item because there was no order.
Here is a video of the applied changes:

https://github.com/user-attachments/assets/e32108f0-b4a1-4300-8ba5-1d062b626273

### Previous Behavior
The old item list did not have any order. This made very time consuming to find any specific item.

### Root Cause
A design flaw. Basic ordering was missing.

### New Behavior
Now the items are ordered alphabetically, allowing to go through the list much faster in order to find an specific item. 

### Fix Implementation
An insertion sort was implemented without including any extra libraries. Items are collected into a temporary list of (name, id) pairs using std::wstring to correctly handle wide strings, sorted alphabetically by name, then used to populate m_itemIds and m_buttonListItems in sorted order.

## Related Issues
None

### DIFF
--- a/Minecraft.Client/Common/UI/UIScene_DebugOverlay.cpp
+++ b/Minecraft.Client/Common/UI/UIScene_DebugOverlay.cpp
@@ -46,7 +46,7 @@ UIScene_DebugOverlay::UIScene_DebugOverlay(int iPad, void *initData, UILayer *pa
     {
         if (Item::items[i] != NULL)
         {
-            sortedItems.emplace_back(std::make_pair(std::wstring(app.GetString(Item::items[i]->getDescriptionId())), i));
+			sortedItems.emplace_back(std::wstring(app.GetString(Item::items[i]->getDescriptionId())), i);
         }
     }
 


Minecraft.Client/Common/UI/UIScene_DebugOverlay.cpp
```diff
@@ -39,17 +39,37 @@ UIScene_DebugOverlay::UIScene_DebugOverlay(int iPad, void *initData, UILayer *pa
 	m_buttonSetNight.init(L"Set Night", eControl_SetNight);
 
 	m_buttonListItems.init(eControl_Items);
+	
+    // Sort items alphabetically
+    std::vector<std::pair<std::wstring, unsigned int>> sortedItems;
+    for (unsigned int i = 0; i < Item::items.length; ++i)
```
void2012: Use size_t for counter variable

void2012 (changes requested):

Minecraft.Client/Common/UI/UIScene_DebugOverlay.cpp
```diff
@@ -39,17 +39,37 @@ UIScene_DebugOverlay::UIScene_DebugOverlay(int iPad, void *initData, UILayer *pa
...
+    {
+        if (Item::items[i] != NULL)
+        {
+            sortedItems.push_back(std::make_pair(std::wstring(app.GetString(Item::items[i]->getDescriptionId())), i));
+        }
+    }
 
+    for (int i = 1; i < (int)sortedItems.size(); ++i)
```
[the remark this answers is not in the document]
CS-DaviMagalhaes: Thanks! I modified it.

void2012:
Record a video how does it work now and I'll merge this PR.

Minecraft.Client/Common/UI/UIScene_DebugOverlay.cpp
```diff
@@ -46,7 +46,7 @@ UIScene_DebugOverlay::UIScene_DebugOverlay(int iPad, void *initData, UILayer *pa
     {
         if (Item::items[i] != NULL)
         {
-            sortedItems.push_back(std::make_pair(std::wstring(app.GetString(Item::items[i]->getDescriptionId())), i));
+            sortedItems.emplace_back(std::make_pair(std::wstring(app.GetString(Item::items[i]->getDescriptionId())), i));
```
void2012: remove std::make_pair too, it's redundant.

CS-DaviMagalhaes: Removed it and tested it, works fine. Thanks again.

Issue #175, opened 2026-03-02
new chunks dont render up to the max render distance


Atomlxe:
https://github.com/user-attachments/assets/2b9b7d4a-1c5a-4768-843e-a93df023744d

old chunks that were already loaded still render up to max distance, but it seems like new chunks only get rendered up to about half that distance

simulpingus:
can confirm this happens

void2012:
This behavior is consistent with the old versions of the Java editions (I assume somewhere before 1.12)

PR #344, opened 2026-03-04
Fix culling that limited render distance of new chunks

ModMaker101:
## Description
Fix frustum culling for newly generated chunks so they render to the full render distance limit instead of being  excluded.

### Previous Behavior
Only chunks marked with `CHUNK_FLAG_COMPILED` had their visibility explicitly set based on the frustum test. Newly generated chunks relied on a separate condition which did not consistently update `pClipChunk->visible in the same way. This caused new chunks to render at a shorter distance than already compiled one.

### Root Cause
`clip` was performed inside the compiled chunk branch meaning visibility logic differed depending on chunk state. Which cased inconsistent visibility updates causing premature culling of new chunks.

### New Behavior
All chunks now undergo the single frustum test at the start of the block.

### Fix Implementation
- Moved the frustum test outside the compiled chunk branch

## Related Issues
#175


ModMaker101:
What does that even mean?

PR #354, opened 2026-03-04
Prevent world input from affecting inventory

### LINKED ISSUE #312: right clicking then opening the creative inventory clears the hotbar
the hotbar gets cleared if you right click and then open the creative inventory, it seems to carry the input over once its opened which should not be happening.

video of the bug:
https://streamable.com/7qng69

Alezito2008:
## Description
Prevents world input from affecting inventory

### Previous Behavior
The hotbar gets cleared if you right click and then open the creative inventory

### Root Cause
The `UIScene_AbstractContainerMenu` was not clearing the input buffer during its construction

### New Behavior
Menus now start with a clean slate. When any container menu is opened, all mouse button states are reset. The player must release and click again to perform any action inside the UI, ensuring no accidental interactions occur during the transition.

### Fix Implementation
Modified the `UIScene_AbstractContainerMenu` constructor implementing a loop that calls `KMInput.ConsumeMousePress(btn)` for the mouse buttons (0, 1, and 2) during initialization.

## Related Issues
- Fixes #312

### DIFF
--- a/Minecraft.Client/Common/UI/UIScene_AbstractContainerMenu.cpp
+++ b/Minecraft.Client/Common/UI/UIScene_AbstractContainerMenu.cpp
@@ -33,6 +33,11 @@ UIScene_AbstractContainerMenu::UIScene_AbstractContainerMenu(int iPad, UILayer *
 	m_bHasMousePosition = false;
 	m_lastMouseX = 0;
 	m_lastMouseY = 0;
+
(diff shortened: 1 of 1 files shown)


Alezito2008:
[Minecraft 2026-03-04 00-00-15 (720p).webm](https://github.com/user-attachments/assets/5807a3a7-407c-42b8-9664-818f0a36e90d)

PR #401, opened 2026-03-04
HighFPS Pause Fix

koreanspy:
## Description
Pressing pause on high frame-rate (200-300+ fps, I'm not sure but the issue isn't as bad on lower fps) causes CANCEL and PAUSEMENU to be occasionally executed at unwanted times. So pressing escape while in the inventory could back you out into the pause menu with a single press.

## Changes
[Video of Before and After](https://share.burritobot.net/03_04_2026/0304_1.mp4)

### Previous Behavior
Press pause once with over 200-300 fps, then have 
Alternatively, open inventory, press escape once to back out, then get sent to the pause menu.

### Root Cause
Obviously, B and pause weren't meant to be bound to the same button, so an additional check needs to be made to prevent pause being executed directly after B.

### New Behavior
Input just checks if an existing UI is open before pushing PAUSEMENU.

### Fix Implementation
Added `!ui.GetMenuDisplayed(i)`
Here -> `if(g_KBMInput.IsKeyPressed(KeyboardMouseInput::KEY_PAUSE) && !ui.IsTutorialVisible(i) && !ui.GetMenuDisplayed(i))`


CeIIHunter:
+1 please merge

PR #410, opened 2026-03-04
Replace CopyAssets.cmake with multithreaded Robocopy (and rsync) implementation.

PortalG:
## Description
`CopyAssets.cmake` has portability/reliability issues that stem from how CMake does file globbing. Rather than trying to find hacky solutions to fix a buggy feature, this PR is an attempt to introduce Robocopy as a reliable, flexible, and multithreaded alternative.

## Changes
1. Scrap `CopyAssets.cmake` completely. (or keep it as an option, your pick)
2. Add two `robocopy.exe` calls in `CMakeLists.txt`: one to copy shared library redists, and one to copy only necessary assets in `Minecraft.Client` using file and folder exclusion patterns.

### Previous Behavior
`CMakeLists.txt` executes `CopyAssets.cmake`, which relies on file globbing to try and find runtime required files to copy over. Not only is CMake's file globbing inefficient, but it is known to cause major portability issues. Depending on what version of CMake you are using or what directory the CMake project is in might affect how files are globbed, causing scenarios where `CopyAssets.cmake` will fail. Runtime is then never reached in a build that is supposed to be automated. When file globbing doesn't work, it often produces output that looks like this:

Screenshot 2026-03-03 163232

Not beneficial for an out-of-box experience.

### Root Cause
This isn't a PR to go and hate on CMake. CMake is a great C/C++ build system compared to others. This is more to address a huge limitation when it comes to copying files the CMake way. Having inconsistent build results for something that is non-obvious and can't be figured out easily by a maintainer isn't great, and should be acknowledged.

### New Behavior
Since the host machine is guaranteed to be building with Windows, Robocopy will always be available as it comes default with Windows systems. With native, multithreaded performance and the flexibility to modify behavior to high degrees, it allows for a more developer controlled experience at the cost of virtually nothing. As a result, the game runs without crash after compiled, with all assets properly copied:

Screenshot 2026-03-03 212029

### Fix Implementation
The `execute_process(...)` function was used to start instances of `robocopy.exe`. As explained before, it uses two calls: one to copy redists and one to copy assets. By default, multithreading is enabled with the `/MT` tag. Other tags are used to ignore empty directories and files that can't be accessed. I tried to implement exclusions as close as the original CMake file, but if I made any mistakes or anything is missing, feel free to add.


void2012:
Please refrain from using Win32-only utilities like that. Or at least hide within `if(...)` that checks that the current system is Windows.

void2012:
@PortalG ready to merge?

Issue #289, opened 2026-03-03
Animation while holding an item


starseed12345:
It would be great if the animations while holding an item like punching or rmb to be adapted to how they look in newer versions as this one looks rather bad and i don't remember it being like this

disepi:
newer swing animations could be taken from the later java decomps, all the required math is already ported in ItemInHandRenderer.cpp just needs a couple different translations

PR #418, opened 2026-03-04
Fix Item First-Person Swing Animation

letsgoawaydev:
## Description

This PR corrects the Swing animation speed.

### Previous Behavior

Previously, the animation would be slower.
Old arm swing animation (video): https://github.com/user-attachments/assets/a4136199-cb9f-4ef7-aac8-6e1aac9d890d

### Root Cause

image
Previously, the `defined _WINDOWS64` was not there, which meant the console edition was using the slower cooldown on windows. The comment states that the reason for it being there on non-hd consoles is due to issues with composite/component signals and fast animations which would cause flickering.

### New Behavior

New arm swing animation (video): https://github.com/user-attachments/assets/b3caba29-a6a2-4b59-9cc2-56614e4b7e0a

### Fix Implementation
Adding `defined _WINDOWS64` ensures that it builds with regular speed for PC.

### Room for Improvement
- Could be made into a toggleable option

## Related Issues
- Related to #289

### DIFF
--- a/Minecraft.Client/ItemInHandRenderer.cpp
+++ b/Minecraft.Client/ItemInHandRenderer.cpp
@@ -548,7 +548,7 @@ void ItemInHandRenderer::render(float a)
         glPushMatrix();
         float d = 0.8f;
 
-#if defined __ORBIS__ || defined __PS3__
+#if defined __ORBIS__ || defined __PS3__ || defined _WINDOWS64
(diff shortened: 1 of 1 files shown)


letsgoawaydev:
I'm pretty sure it was faster on xbox one too but i could be wrong

letsgoawaydev:
theres always room for improvement, i think that there should be a toggle for this fix for the future

PR #423, opened 2026-03-04
Add Control Remapping & Modernize Options Persistence

wize-logic:
Screenshot 2026-03-04 104806
vpbsc7
Screenshot 2026-03-04 192532

## Description

This PR adds a complete control remapping system for both keyboard and controller inputs, adds TOML-based options persistence (via toml11) with a companion JSON file, and moves the options file to the executable root directory.

### Previous Behavior

- **No control remapping UI existed** — keyboard bindings were hardcoded to WASD/mouse defaults with no way for the user to change them in-game. Controller bindings used a fixed layout with no remapping capability.
- **No structured options persistence** — the options format had no support for nested data like input mapping tables.
- **Keyboard input was hardcoded** — `Input.cpp` used `KEY_W`, `KEY_A`, `KEY_S`, `KEY_D` directly. `Minecraft.cpp` used hardcoded `VK_` codes for attack, use, jump, sneak, drop, and inventory actions.
- **The `Keyboard` stub class** had a limited set of key constants (A-Z, Space, Shift, Escape, Backspace, Enter, arrows, Tab) with no display name support, no virtual key conversion, and no reverse-lookup capability.

### Root Cause

- The codebase is for a console game (Xbox 360, PS3, etc.) where controller layouts were fixed. The Windows 64-bit port inherited this design with hardcoded keyboard defaults.
- The options format couldn't represent structured data like key binding tables.

### New Behavior

- **Full control remapping** — users can rebind all 14 keyboard actions and 13 controller actions through an in-game UI. Two implementations exist:
  - `UIScene_ControlRemapMenu` (console-style, uses JoinMenu SWF with ButtonList + info labels)
  - `ControlRemapScreen` (PC-style, uses Screen class with SmallButton widgets)
- **Input detection dialog** — `UIScene_InputPrompt` presents a "Press a key/button..." overlay that detects keyboard keys, mouse buttons, or controller buttons via polling.
- **TOML + JSON dual persistence** — settings are saved to both `options.toml` (primary, human-readable) and `options.json` (companion) simultaneously in the exe root directory. The TOML format uses human-readable names for keys (`"W"`, `"SPACE"`, `"L.MOUSE"`) and controller buttons (`"RT"`, `"A"`, `"LB"`).
- **Backward compatibility** — loading supports old integer key codes and numeric controller bitmask formats, auto-migrating to the new format on save.
- **All keyboard input is now remappable** — `Input.cpp` and `Minecraft.cpp` read from `Options::keyMappings[]` instead of hardcoded constants. Helper functions `remapConsumePress()` and `remapIsDown()` handle the mouse-button negative encoding transparently.
- **Extended `Keyboard` class** — 51 key constants (added arrow keys, number keys, Ctrl, Alt), `getKeyName()` for display, `keyToVK()` for Windows virtual key conversion, `getKeyCount()` for enumeration.

#### Control Remapping UI

1. **`UIScene_ControlRemapMenu`** — Reuses the `JoinMenu` SWF (the only available SWF with a Button + ButtonList + label pairs). Maps `JoinGame` button to "Reset All Defaults", `GamePlayers` ButtonList to 4 action items (Previous/Next/Set Keyboard/Set Controller), and `Label0-2`/`Value0-2` to display current action name, keyboard key, and controller button.

2. **`UIScene_InputPrompt`** — Reuses the `MessageBox` SWF. When opened, waits for all inputs to be released, then polls either keyboard (all 51 keys + 3 mouse buttons via `Keyboard::isKeyDown()` / `Mouse::isButtonDown()`) or controller (XInput digital buttons + analog triggers with threshold > 128). Returns detected value via C-style callback.

3. **Scene registration** — Added `eUIScene_ControlRemapMenu` and `eUIScene_InputPrompt` to `EUIScene` enum in `UIEnums.h`, added factory cases in `UILayer::createScene()`, added navigation from `UIScene_HelpAndOptionsMenu` and `OptionsScreen`.

#### Input Remapping Integration

4. **`Input.cpp`** — Replaced hardcoded `KEY_W/A/S/D` with `Keyboard::keyToVK(opts->keyUp/Down/Left/Right->key)`. Added keyboard jump binding.

5. **`Minecraft.cpp`** — Added `remapConsumePress(key)` and `remapIsDown(key)` helpers that handle both positive key codes (via `keyToVK`) and negative mouse button codes (via `key + 100`). Rewired attack, use, jump, sneak, drop, inventory, and chat actions to use `Options::keyMappings[]`.

6. **`Options::applyControllerMappings()`** — Maps the 13 controller action indices to `MINECRAFT_ACTION_*` enums and calls `InputManager.SetGameJoypadMaps(MAP_STYLE_0, ...)` for each.

#### TOML Persistence (toml11)

7. **Added TOML support via toml11 v4.4.0** — Integrated the toml11 single-header library (C++11 compatible) at `Minecraft.Client/toml11.hpp` to enable structured options persistence. Uses exception-less parsing (`toml::try_parse_str()`), typed value extraction (`toml::find_or<T>()`), and table construction via `toml::value(toml::table{...})` initializer-list syntax.

8. **`Options::load()`** — Tries `options.toml` first (parsed via `toml::try_parse_str()`), falls back to `options.json` (parsed via simple `jsonExtractValue()` string extractors), creates defaults if neither exists. Keyboard section tries string values then integer fallback. Controller section tries human-readable action/button names then numeric index/bitmask fallback.

9. **`Options::save()`** — Builds a `toml::value` with all settings using initializer-list syntax, serializes via `toml::format()`. Also writes a companion `options.json` with the same data. Deletes existing files before writing to avoid stale trailing data.

10. **Name resolution** — `getKeyByName()` reverse-maps display names (`"W"` -> `KEY_W`, `"L.MOUSE"` -> `-100`). `getControllerButtonByName()` reverse-maps button names (`"RT"` -> `0x00400000`).


Boom244:
This is pretty good. I don't have an Xbox 360 controller lying around and I can't for the life of me seem to get x360CE working with my DualShock 4, so I couldn't test whether controller key binds were functional yet. Either way, some things to consider:

- Not sure if this is just me, but key binding does not seem to recognize many non-alphanumeric keys (i.e., `,`,`.`,`;`). I think this will probably be an issue in the future, especially when it comes to people wanting to bind up special buttons on their gaming keyboards or mice or whatever. 

- In Java Edition, key bind remapping will cause colliding keys to go red  (and Java Edition allows you to scroll cleanly between key binds, so you can see this easily). In your patch, colliding keys don't seem to have any sort of indication, which becomes a problem when I force a collision (i.e., binding both Inventory and Drop to E, as I have in the screenshot below, which makes Inventory take precedence and dropping impossible.)
image

(The next two are _really_ nitpicky on my part, ignore if you don't think it's worth listening to) 

-  While I do appreciate the effort you went to to build your own menu for control remapping, as my own stylistic opinion, the control schema is kind of... cumbersome? Rebinding shouldn't make me have to go halfway around the world for one specific key. I suggest refactoring, if it's not too much trouble, into a menu more aligned with the style of either Java Edition's controls menu or the Start/Join Game menu, just to have that list feel that many PC players are used to.

- We should have a way to unbind certain actions. I can see that, in the far future, it would be useful to be able to unbind unused actions in the event of someone adding a mod or multiple mods that take up keyboard space with custom actions, as they do often in Java Edition.

PR #432, opened 2026-03-04
Fix incorrect distance comparison return value #417

### LINKED ISSUE #417: Wolf attack crash
Game crashes when wolves attack mobs

https://imgur.com/a/YQYmeeE

ModMaker101:
## Description
Fixes an incorrect return value in the distance comparator used by `NearestAttackableTargetGoal`, which caused crashes during wolf target selection.

### Previous Behavior
When two distances were equal the comparator returned true, breaking the weak ordering required by STL containers and causing wolf attack crashes.

### Root Cause
The comparator treated equal distances as '<`.

### New Behavior
Equal distance now correctly return false.

### Fix Implementation
Replaced the final `return true;` with `return false;` in `DistComp::operator()`.

## Related Issues
- Fixes #417


ModMaker101:
Ready to merge.

PR #446, opened 2026-03-04
Persist fullscreen setting and start windowed mode maximized

### LINKED ISSUE #391: Game should have run in fullscreen mode on next game launch while fullscreen is already toggled!
When you toggle fullscreen in Minecraft.Client and trying to exit game and relaunch, the game runs on Windowed mode instead of fullscreen mode

MCbabel:
## Description
Persist the fullscreen toggle state across game restarts and start the window maximized when in windowed mode.

### Previous Behavior
- Toggling fullscreen with F11 only lasted for the current session. On the next launch, the game always started in windowed mode regardless of the user's last fullscreen setting.
- When starting in windowed mode, the window opened at a non-maximized size, requiring the user to manually maximize it.

### Root Cause
`g_isFullscreen` was hardcoded to `false` on line 98 of `Windows64_Minecraft.cpp` and was never saved to disk. The `ToggleFullscreen()` function only modified the in-memory variable without any persistence. There was also no mechanism to read a saved fullscreen preference on startup.

### New Behavior
- The fullscreen state is saved to `options.txt` (next to the executable) whenever F11 is pressed. On the next launch, the saved state is read and applied after window creation.
- When the game starts in windowed mode, the window is now shown maximized by default.

### Fix Implementation
- Added `SaveFullscreenOption()` and `LoadFullscreenOption()` helper functions that read/write a simple `fullscreen=1` or `fullscreen=0` key-value pair to `options.txt` (following the same pattern as the existing `username.txt`).
- `ToggleFullscreen()` now calls `SaveFullscreenOption()` after toggling the state.
- In `_tWinMain()`, after `InitDevice()` completes, `LoadFullscreenOption()` is called and `ToggleFullscreen()` is invoked if the saved state is fullscreen.
- Changed `ShowWindow()` in `InitInstance()` to use `SW_SHOWMAXIMIZED` instead of the default `nCmdShow` for non-hidden windows.

## Related Issues
- Fixes #391

### DIFF
--- a/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
+++ b/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
@@ -124,6 +124,50 @@ static void CopyWideArgToAnsi(LPCWSTR source, char* dest, size_t destSize)
 	dest[destSize - 1] = 0;
 }
 
+// ---------- Persistent options (options.txt next to exe) ----------
+static void GetOptionsFilePath(char *out, size_t outSize)
+{
+	GetModuleFileNameA(NULL, out, (DWORD)outSize);
+	char *p = strrchr(out, '\\');
+	if (p) *(p + 1) = '\0';
+	strncat_s(out, outSize, "options.txt", _TRUNCATE);
+}
+
+static void SaveFullscreenOption(bool fullscreen)
+{
+	char path[MAX_PATH];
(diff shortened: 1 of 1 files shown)


void2012:
Related to #320 ? Please coordinate with that PR author if so.

void2012:
So I should merge yours?

EnderActually (approved):
Works good on the Linux side!

PR #460, opened 2026-03-04
Implement Crossplatform (Linux, macOS, etc) support

izzint:
This is a major overhaul to the current way the project is structured, as well as an entirely new backend (`Cross64`) intended on implementing a more modern and non-proprietary toolset for said backend.

**This branch is currently _not finished_ and this branch is not currently able to be compiled** into an actual native Linux, macOS or whatever newfangled *nix the kids are using these days binary, however I see that changing very verryy soon ;)

### DIFF
--- a/Minecraft.Client/Cross64/Cross64Extras/Cross64Stubs.h
+++ b/Minecraft.Client/Cross64/Cross64Extras/Cross64Stubs.h
@@ -32,15 +32,15 @@ inline VOID GetSystemTime(LPSYSTEMTIME lpSystemTime) {
   const hh_mm_ss hms{dateTime - dp};
   const weekday wd{dp};
 
-  lpSystemTime->wYear = ymd.year();
-  lpSystemTime->wMonth = ymd.month();
-  lpSystemTime->wDayOfWeek = wd.c_encoding(); // 0 = sunday, etc
-  lpSystemTime->wHour = hms.hours();
-  lpSystemTime->wMinute = hms.minutes();
-  lpSystemTime->wSecond = hms.seconds();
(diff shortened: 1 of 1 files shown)


void2012:
#465 Does it cleaner though

void2012:
> Native Linux, MacOS binaries

Do you realize what kind of a herculian task it would be? WinAPI, Iggy, Miles, DirectX11 to name a few... We are too far from this and this is not the highest priority now. The amount of commit conflicts you will have to deal with every single commit (entirely yourself) will be overwhelming.

First, remove the entire `#include`s thing because #465 does it cleaner and it will make this PR (if it's actually serious) easier to review because maintainers can concentrate on the code itself.

izzint:
Someone has already made a stub for miles that replaces it with miniaudio and it works perfectly, some smaller parts (DirectX, Win32 stubs) are very easy to stub out (and have been before, look at the PS4 and PS3 ports, that also use custom renderers)

izzint:
> Vulkan renderer is far more practical and realistic to implement now so that is welcome(as an alternative renderer of course, not many GPU and compatibility layers support it). "Cross-platformness" is multi-layered problem and shouldn't be done in a large fat PR. I'm closing this one for good.

What would be the recommended path to implementing this without building another backend folder?

DecalOverdose:
also what do you mean by "In the spirit of wine", and it is how you port software, and it appears that you use Windows so.... :clown_face:

MasterBroNetwork:
This is such unnecessary drama to have for a game that is over a decade old, running off of leaked source code.

PR #465, opened 2026-03-04
Support for cross-compiling on Linux

universeindex:
## Description
I've implemented support for cross-compiling a Windows build on a Linux host using clang-cl. I added a toolchain file at cmake/LinuxCrosscompile.cmake that achieves this. I also modified many source files to not use backslashes in include statements (which error with clang-cl) and use case-sensitive paths (many Linux filesystems are case-sensitive, unlike NTFS, which causes errors). I avoided changing actual functionality as much as possible, so I hope this isn't too intrusive. Additionally I apologize for the messy commit messages, I hope this isn't a big issue, I have detailed all of the changes I made here.

## Changes
Include statements use forward slashes instead of backslashes
Include statements reference files with correct capitalization
Added a CMake toolchain file for using clang-cl on Linux
Minecraft.Client/Xbox/MinecraftWindows.rc is not compiled on Linux, this causes errors otherwise, as it can't find the windows SDK and I am unsure how to specify the location here.
I also added documentation for building on Linux.

### Previous Behavior
Running CMake on Linux would throw a hardcoded error stating that Linux isn't supported. When removed, and clang-cl is set to be used, many errors occur due to include paths using backslashes and inconsistent capitalization. The only way to build and run on Linux before is to use a Windows VM, and there is no way to get proper code completion on a Linux IDE.

### Root Cause
Backslashes in include paths are supported on Windows, but not other systems, and NTFS is case-insensitive, so include paths do not have to match capitalization, also not on other systems.

### New Behavior
The game now compiles and outputs a Windows binary on Linux using clang-cl.

### Fix Implementation
See changes section


void2012:
Have you tested it compiling with MSVC? Both msbuild and cmake. I hope this doesn't break the MSVC compilation.

cmake/LinuxCrosscompile.cmake
```diff
@@ -0,0 +1,18 @@
+set(CMAKE_SYSTEM_NAME Windows)
+
+set(CMAKE_C_COMPILER clang-cl)
+set(CMAKE_CXX_COMPILER clang-cl)
+set(CMAKE_LINKER lld-link)
+
+set(CMAKE_C_COMPILER_FRONTEND_VARIANT MSVC)
+set(CMAKE_CXX_COMPILER_FRONTEND_VARIANT MSVC)
+
+add_compile_options(/winsysroot /opt/msvc --target=x86_64-pc-windows-msvc -fms-compatibility -fms-extensions -fdelayed-template-parsing)
```
DwifteJB: the cmake file assumes /opt/msvc whereas https://github.com/mstorsjo/msvc-wine recommends ~/my_msvc/opt/msvc

just a simple issue as you can syslink or run the msvc-wine installer as root & install to /opt/msvc

DwifteJB:
> > > I think `__iob_func` should be implemented as `__declspec(naked)` + `__asm`.
> > 
> > 
> > `_declspec(naked)` seems to not be working since `__asm` seems to have been dropped for x64 (if you want to use it you have to use a .asm file, like how 4J seems to have done it)
> > `__attribute__((naked))` & `__asm__()` works however
> 
> Forgot about it that, as I am used to x86 development. Uncool, Microsoft. Anyway, maybe conditionally use the `.asm` file for MSVC assembler and `.cpp` with `[[__gnu__::__naked__]]` for Clang-cl?

yeah i havent done much c++ for a good few years and thought it still existed,

this seems to work though https://github.com/DwifteJB/MinecraftConsoles/blob/ee78a74025b78b0c5c813a7beb7c94629a30c2ff/Minecraft.Client/IobShim.cpp

universeindex:
@DwifteJB merged your pr, thx

DwifteJB:
> so, the changes to the LinuxCrosscompile.cmake actually broke completely, it never detects the windows sdk location correctly and falls back to /opt/msvc
> 
> I never noticed this because that's actually where mine is installed, but someone in our matrix server was having this issue building even on a clean debian system
> 
> as such I rewrote it to require manually specifying the location with -DWINDOWS_SDK_PATH

hmm that's super strange, maybe my install is just a bit different from yours? i mean i only really tested i with msvc-wine rather than a manual install of the visual studio build files

we could always use a wrapper like build.sh to find a using whereis since i know clang & cmake has some issues with find_package (especially when running as root)

DwifteJB:
looks good but would be nice to get that pr i made pushed into this pr (or i can just make a seperate one after you can merge)

void2012:
okay I'll wait

void2012 (changes requested):

universeindex:
linux cmake is working, but if you could do windows cmake that would be greatly appreciated (with both msvc and clang-cl)

universeindex:
I use neovim and clangd

universeindex:
now the solution isn't even opening
I'm about to firebomb microsoft HQ

DwifteJB:
via cmake:
```
4J_Render_PC_d.lib(pngerror.obj) : error LNK2019: unresolved external symbol __iob_func referenced in function png_defa
ult_error [C:\Users\builds\MinecraftConsoles\build\MinecraftClient.vcxproj]
C:\Users\builds\MinecraftConsoles\build\Debug\MinecraftClient.exe : fatal error LNK1120: 1 unresolved externals [C:\Use
rs\builds\MinecraftConsoles\build\MinecraftClient.vcxproj]
```

DwifteJB:
> > via cmake:
> > ```
> > 4J_Render_PC_d.lib(pngerror.obj) : error LNK2019: unresolved external symbol __iob_func referenced in function png_defa
> > ult_error [C:\Users\builds\MinecraftConsoles\build\MinecraftClient.vcxproj]
> > C:\Users\builds\MinecraftConsoles\build\Debug\MinecraftClient.exe : fatal error LNK1120: 1 unresolved externals [C:\Use
> > rs\builds\MinecraftConsoles\build\MinecraftClient.vcxproj]
> > ```
> 
> was this with msvc or clang, can you check if this line is in ClientSources.cmake? `$<$<CXX_COMPILER_ID:MSVC>:iob_shim.asm>`

msvc and there is   $<$<CXX_COMPILER_ID:MSVC>:iob_shim.asm>
  $<$<CXX_COMPILER_ID:Clang>:IobShim.cpp>

universeindex:
image

universeindex:
also slightly unrelated but i did notice instruction set extensions were disabled, it might be worth enabling SSE2 at least on windows64 for instance
it was first released in 2000 so any system released since then should ideally support it
there is also avx released in 2011, or avx2 in 2013

void2012:
> also slightly unrelated but i did notice instruction set extensions were disabled, it might be worth enabling SSE2 at least on windows64 for instance it was first released in 2000 so any system released since then should ideally support it there is also avx released in 2011, or avx2 in 2013

SSE2 is enabled by default for all C++ compiler targeting x64. This is because the first x64 CPU (amd64) had SSE2 already. 
As for AVX2 (Steam survey says nearly 90% computers running Steam has AVX2 as of late 2025), I'm considering to add clang-cl compiled binary with AVX2 enabled, ship as an alternative binary for nightly builds and with `/fp:fast`. Though it needs testing.

universeindex:
i've been trying for ages and I genuinely have no clue how to add a clang build option to msbuild, really sorry. it works fine in cmake but visual studio just isn't playing along. if someone else can figure it out thats great, but I give up

universeindex:
I get that, but my issue is that it's just not creating a new configuration. I select new and click ok then nothing happens, doesn't add the config at all
https://files.catbox.moe/ds1lvy.mp4

DwifteJB:
> > ... this is what this pull request is doing... 😭😭😭
> 
> 
> 
> Is there a script included that already does this? If so then ignore me I guess lol
> 
> I just saw people discussing llvm-rc which isn't actually needed if done this way.

there's no script to install but it's mentioned in compile.md within the project, might be a good idea to use something like docker / nix so we can containerise the actual building so it works no matter what system (macos, linux, windows, etc)

niansa:
There are Dockerfiles already in the repo lol

Either way in case anyone can use it, here's a script I wrote for testing purposes:
```bash
# Fresh Debian Trixie install

dpkg --add-architecture i386
apt update
apt install wine wine32 wine64 msitools git python3 cmake ninja-build lld-19 llvm-19 clang-19

DISPLAY= wineboot

# MSVC install
(
    git clone https://github.com/mstorsjo/msvc-wine.git
    cd msvc-wine
    git checkout 32b504c63b869681cda6824a20e30b74cb718432

    mkdir /opt/msvc
    ./vsdownload.py --dest /opt/msvc --accept-license
    ./install.sh /opt/msvc

    ln -s /usr/bin/clang-cl{-19,}
    ln -s /usr/bin/lld-link{-19,}
)

# MSVC setup
export PATH="/opt/msvc/bin/x64:$PATH"

# Project build
(
    git clone https://github.com/universeindex/MinecraftConsoles.git --depth 1 --recursive
    cd MinecraftConsoles
    mkdir build
    cd build
    cmake .. -GNinja -DCMAKE_TOOLCHAIN_FILE=../cmake/LinuxCrosscompile.cmake
    ninja
)
```

BlueishSapphire:
Would it be reasonable to split this into multiple PRs? Converting backslashes to forward slashes and fixing the casing issues in `#include`s would be easy to split off into a new PR and merge it right now. It would also make developing on HEAD on linux a lot easier in the meantime while the rest of this PR is still being developed.

universeindex:
We love microslop <3
I'll see if i can fix merge conflicts soon but I am still a little busy at school right now, so sorry if it takes a few hours

DwifteJB:
seems like it, ill check to see whats going on with it now by running the actions locally

Twig6943:
> so im just making it run on ubuntu itself rather than an arch linux container within it...

Could be the case ig? I tried fedora and it also got stuck at the same place

DwifteJB:
@universeindex https://github.com/universeindex/MinecraftConsoles/pull/4 very simple fix inside

Twig6943:
@universeindex yo?

Twig6943:
@codeHusky could you close this as the other pr has been merged?

codeHusky:
Superceded by #1403

PR #478, opened 2026-03-04
Add servers.txt so players can add an arbitrary amount of servers to the "Join Game" list

Siepert123:
## Description
Allows players to define server entries in `servers.txt` (located in the folder of the running EXE)

## Changes
Added code that reads `servers.txt` and adds defined servers to the list
The text file is defined as follows:
```
<ip>
<port>
<display name>
```
and it can have this as many times as one wants.
Example `servers.txt` file:
```
155.212.208.147
25565
Dryk server!
26.31.243.15
25565
Sus server
127.0.0.1
25565
Lokale host tm
```
This produces the following result:
image

### Previous Behavior

Previously, the join game list only displayed local LAN connectables with no way for the user to direct connect to other servers.

### Root Cause

No input for the player to add servers

### New Behavior

The player can define server entries in `servers.txt` and they will be added to the Join Game list

**IMPORTANT: I added `#ifdef _WINDOWS64` as I am not sure how this behaves on other platforms!**
Please act accordingly.

### Fix Implementation

First, the game checks for the text file, and if it exists, iterates over each line parsing it as `ip`-`port`-`name` until the file ends
Everytime `name` is parsed it adds the information to the list!

## Related Issues
There are no related issues i think

### DIFF
--- a/Minecraft.Client/Common/Network/PlatformNetworkManagerStub.cpp
+++ b/Minecraft.Client/Common/Network/PlatformNetworkManagerStub.cpp
@@ -701,6 +701,7 @@ void CPlatformNetworkManagerStub::SearchForGames()
 #ifdef _WINDOWS64
 	std::vector<Win64LANSession> lanSessions = WinsockNetLayer::GetDiscoveredSessions();
 
+	//THEY GET DELETED HERE DAMMIT
 	for (size_t i = 0; i < friendsSessions[0].size(); i++)
 		delete friendsSessions[0][i];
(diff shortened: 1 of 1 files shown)


DrChopSticks:
This! is so much more user friendly for now than launch params, at least until a custom ui can get implemented for this. and even then you could just piggy back of this code for the custom ui eventually.

DrChopSticks:
> how do I even install this?

you have to wait until its merged into the main branch

PR #481, opened 2026-03-04
Fix gamma slider via pixel shader #178

### LINKED ISSUE #178: Gamma slider has no effect

ModMaker101:
## Description
This PR fixes the gamma slider so it properly affects scene brightness on all platforms. Instead of relying on DXGI system level gamma, gamma is now applied directly through the light-texture pipeline

### Previous Behavior
- Gamma slider did almost nothing.
- Originally attempted hardware gamma via `IDXGIOutput::SetGammaControl`, which was unreliable.  

### Root Cause
- The engine never applied gamma in a sofware path.
- Hardware gamma APIs were inactive in the environments people were using.

### New Behavior
- Gamma slider now affects global brightness consistently.
- A small toe lift keeps shadows readable when brightend.
- Works per-player in split-screen.

### Fix Implementation
- Removed reliance on DXGI.
- Added software gamma curve inside `updateLightTexture`, with clamping and toe correction
- Verified in day/night, caves, underwater, and rain.

## Related Issues
- Fixes #178 

Before:
image

After:
image

Not entirely too much of a difference during the day.

Before:
image

After:
image

Now I'm WAY to lazy to try and time it up to get a lighting strike, but yeah the gamma bar also works too:

https://github.com/user-attachments/assets/35f85595-a0d3-4ab9-a73d-ed367684c20e


void2012:
goated PR, thanks

void2012:
@ModMaker101

ModMaker101:
@void2012 ready for review. One caveat is I kept getting an exception for multiplayer thing which is in progress so I wasn't able to test split-screen.

image

PR #531, opened 2026-03-05
Fixes Alt+Enter fullscreen

### LINKED ISSUE #499: Fullscreen status only persists if fullscreened with F11, not alt+enter
Fullscreen persistence only works if the game was fullscreened using the F11 key, alt+enter also fullscreens the game but will not persist on next launch

KoopaCode:
## Description
Fixes fullscreen save so both `F11` and `Alt+Enter` work the same.

## Changes

- `Alt+Enter` now saves fullscreen state the same way `F11` does.
- Fullscreen now comes back on next launch no matter which key combo was used.

### Previous Behavior
Fullscreen only saved correctly when using `F11`.

If you used `Alt+Enter`, it went fullscreen for that session, but it did not stay fullscreen next time you launched.

### Root Cause
There were two different fullscreen paths.

`F11` saved the setting, but `Alt+Enter` did not save.

### New Behavior
Fullscreen now saves and restores correctly for both `F11` and `Alt+Enter`.

### Fix Implementation
Made `Alt+Enter` use the same save path as `F11`.

- Both toggle methods now go through the same fullscreen update logic.
- Fullscreen state is written when either method is used.
- No other fullscreen behavior was changed.

## Related Issues
- Fixes #499

### DIFF
--- a/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
+++ b/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
@@ -96,6 +96,44 @@ wchar_t g_Win64UsernameW[17] = { 0 };
 // Fullscreen toggle state
 static bool g_isFullscreen = false;
 static WINDOWPLACEMENT g_wpPrev = { sizeof(g_wpPrev) };
+extern HWND g_hWnd;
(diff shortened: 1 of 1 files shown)


mattsumi:
Please check out your merge conflicts

PR #498, opened 2026-03-04
Implementation of Dedicated Server in Isolation

kuwacom:
## Description
This PR introduces a Windows64 dedicated server executable (`Minecraft.Server`) and expands the dedicated-server runtime/build pipeline.

Compared to the initial pull request description, the following features have been newly added:
- interactive server console (linenoise + command completion),
- broader `server.properties` support (including LAN advertise and host options),
- world bootstrap/load behavior based on `level-id`,
- Docker (Wine) runtime support and helper scripts.

`Minecraft.Client` changes remain focused on dedicated-server bridge compatibility.
#65

image

### Dedicated Server Executable / Startup
- Added dedicated server entrypoint and project:
  - `Minecraft.Server/Windows64/ServerMain.cpp`
  - `Minecraft.Server/Minecraft.Server.vcxproj` (+ filters)
- Dedicated startup flow now:
  - loads `server.properties` defaults,
  - applies CLI overrides,
  - bootstraps world by `level-id` / `level-name`,
  - performs initial save for newly created worlds,
  - runs periodic autosave and graceful shutdown save.
- Dedicated launch options:
  - `-port`, `-ip`/`-bind`, `-name`, `-maxplayers`, `-seed`, `-loglevel`, `-help`
  - Since the arguments can basically override `server.properties`, it can also be integrated into fully autonomous deployment systems like Kubernetes.

### Dedicated Runtime Modules
- Added/expanded:
  - `ServerProperties` (defaulting, normalization, persistence for dedicated options)
  - `WorldManager` (load existing world by `level-id`, fallback by name, or create new world)
  - `ServerLogger` (`debug|info|warn|error`)
- Added shared server string utility layer:
  - `Minecraft.Server/Common/StringUtils`
  - used by logger/properties/world/console paths to avoid duplicated conversion/normalize logic.

### Interactive Server Console
- Added console runtime:
  - `ServerCli`, `ServerCliInput`, `ServerCliParser`, `ServerCliEngine`, `ServerCliRegistry`
  - linenoise-based input/history/completion support.
- Added built-in commands:
  - `help` (`?`)
  - `stop`
  - `list`
  - `tp` (`teleport`)
  - `gamemode` (`gm`)
- Added completion behavior:
  - command/alias completion,
  - player name completion,
  - gamemode token completion (`survival|creative|s|c|0|1`).

### Minecraft.Client Bridge Changes (Dedicated Compatibility Scope)
Only dedicated-server compatibility behavior was changed in client networking paths:

[...]

PR #536, opened 2026-03-05
Fix multiplayer player data mix between different players bug

### LINKED ISSUE #509: [MULTIPLAYER BUG] Inventory data may mix between different players
I don't know what's the cause and what triggers it but it happens

kuwacom:
## Description
This PR fixes a Win64 multiplayer identity bug where player save data (`players/*.dat`) could be loaded for the wrong player when join order changed.

On Win64, player identity was effectively tied to connection-order `smallId`-based XUIDs.  
This PR introduces a deterministic username-derived persistent XUID and wires it into the existing XUID-based save/load flow.

https://github.com/user-attachments/assets/c1202d04-9c2e-47ba-a38d-20f08d3533f5

### Previous Behavior
inventory/progress/map ownership could appear swapped between players.

### Root Cause
- Save/load and map ownership logic are keyed by `Player::xuid`.
- Win64 networking identity (`IQNetPlayer::GetXuid`) was tied to `smallId`, not a stable per-user identifier.
- This created instability in player-data lookup whenever connection ordering changed.

### New Behavior
- Added `Windows64_NameXuid` for deterministic `name -> persistent xuid` resolution.
- On Win64 login (`PlayerList`), `ServerPlayer::xuid` is now resolved from username.
- Local player `xuid` assignment (`Minecraft`) is aligned to the same resolver in create/init/respawn paths.
- Added a Win64 local-self guard in `ClientConnection::handleAddPlayer` (name check) to prevent duplicate local remote-player creation.
- `IQNet::GetPlayerByXuid` keeps compatibility fallback behavior, but now also resolves username-based XUIDs.
- Canonical implementation is in `Minecraft.Client/Windows64/Windows64_NameXuid.h`; legacy `Win64NameXuid.h` remains as a compatibility include.

Rename migration is intentionally out of scope (same-name identity only).

## Related Issues
- Fixes #509

### DIFF
--- a/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
+++ b/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
@@ -42,6 +42,7 @@
 #include "..\..\Minecraft.World\OldChunkStorage.h"
 #include "Common/PostProcesser.h"
 #include "Network\WinsockNetLayer.h"
+#include "Windows64_Xuid.h"
 
 #include "Xbox/resource.h"
 
@@ -1220,6 +1221,12 @@ int APIENTRY _tWinMain(_In_ HINSTANCE hInstance,
 	// Load stuff from launch options, including username
 	Win64LaunchOptions launchOptions = ParseLaunchOptions();
 	ApplyScreenMode(launchOptions.screenMode);
+
+	// Ensure uid.dat exists from startup in client mode (before any multiplayer/login path).
+	if (!launchOptions.serverMode)
+	{
+		Win64Xuid::ResolvePersistentXuid();
+	}
 
 	// If no username, let's fall back
 	if (g_Win64Username[0] == 0)


kuwacom:
OK, everything is working smoothly.
for now, it's complete.

- The XUID is stored as `uid.dat` in the same directory as the executable.
- In multiplayer, the host now also checks for duplicate players by XUID in addition to the user name.
- Compatible with unpatched clients as well.
- In singleplayer (world host mode), the previous embedded XUID is used to maintain compatibility with existing worlds.

codeHusky:
separate issues but still notable (prob split into an issue after this pr)

- reconnecting more than once or twice to a server breaks your ability to connect to it again
  - seems to break connecting for all future players for some reason
- worlds do not save, even when using /save-all or /stop the world data does not save/load properly for players
- jumping out of a server from the server list results in a "texture pack not installed" error on the client for some reason

occurs with dedicated server

kuwacom:
> If this implementation is going to be merged, I’d like to merge it into the dedicated server implementation as well.

The merge I mentioned here refers to the isolation dedicated server in my pull request, not the `-server` dedicated server.
Sorry for the confusion.

I haven’t looked closely at the `-server` implementation, but it should probably be fine to merge it as-is.

kuwacom:
> Where does the isolated dedicated server build to? I only see Minecraft.Client.exe in the builds I'm making in VS

I should have linked it directly from the start.
sorry

#498

kuwacom:
> Yep it's on connect - it'd be good to make the uid.dat generate on game launch and not just on connection to the world. If you change that up it's good to merge, and yes I do intend to merge this in

At startup, I added a check for `uid.dat` in `Windows64_Minecraft.cpp` if it's not running in server mode (i.e., not started with `-server`).

kuwacom:
> Could you merge the changes from #498 into this then? Would make it easier for me to test and it's a similarly important feature

Once this pull request is merged, I’ll continue the remaining work in the #498

Issue #328, opened 2026-03-03
Game softlocks when you attempt to delete a save from here - doesn't respond the neither controller inputs or keyboard inputs - forced to close the game and restart.


rynninglol:
<img width="1715" height="966" alt="Image" src="https://github.com/user-attachments/assets/6269850a-c36e-4813-8bde-12d725ee207d" />

From this section here. Won't respond to anything, have to hit ALT + F4 to force close the game.

PR #539, opened 2026-03-05
Fix save list, delete save, exit without saving, and blank username on Windows64

### LINKED ISSUE #463: Cannot rename worlds in the world creation menu
After KBM rework you can't select the world name in the world creation screen to change the name of the world.

dtentiion:
## Description
Fixes world save renaming, persistent username display, new-world orphan folders, save deletion, and native in-game text input on Windows 64-bit. The rename dialog (Save Options > Rename Save) would open, accept input, and close without applying any change. The typed name was being discarded silently. Additionally, after joining a world, saving, exiting, and re-joining, the player's in-game username would appear completely blank. Creating a new world and immediately choosing "Exit Without Saving" would leave an orphaned save folder behind, causing the unsaved world to appear permanently in the save list. Attempting to delete a save from Save Options had no effect. Text input for world naming used a native Windows API popup dialog instead of an in-game UI, and controller users had no on-screen keyboard at all.

https://github.com/user-attachments/assets/fa62aa32-971a-4a83-91e0-1361f2201e93

https://github.com/user-attachments/assets/33481640-032e-4d7d-b619-a33d12116cd7

https://github.com/user-attachments/assets/a1bccf7e-66fa-4fea-b2d7-d4a56c250842

### Previous Behavior
Selecting a save, opening Save Options, choosing Rename Save, and typing a new name had no effect. The save slot would continue displaying the original name after the dialog closed. After joining a world, saving, and re-joining, the player's username would appear blank on the nametag. Creating a new world and immediately choosing "Exit Without Saving" would leave an orphaned save folder in `Windows64\GameHDD\`, causing the unsaved world to appear permanently in the save list. Choosing "Delete Save" from Save Options would display the confirmation dialog but not remove the save from disk or from the list. Naming or renaming a world used a bare Windows API `InputBox` dialog that was visually inconsistent with the rest of the game. Controller users received no keyboard at all when trying to name a world.

### Root Cause
`KeyboardCompleteWorldNameCallback` in `UIScene_LoadOrJoinMenu.cpp` only called `StorageManager.RenameSaveData()` inside a `#if (defined __PS3__ || defined __ORBIS__ || defined _DURANGO || defined(__PSVITA__))` block. There was no `_WINDOWS64` branch at all, so the new name was validated and then silently thrown away on Windows. For the blank username, `IQNet::EndGame()` in `Extrax64Stubs.cpp` zeroed `m_player[0].m_gamertag` as part of resetting network state on game exit. When `FakeLocalPlayerJoined` was called for a subsequent session, it passed that empty gamertag into the player creation chain, producing a blank in-game name. For the orphan folder, `MinecraftServer.cpp` unconditionally called `saveToDisc` during new-world creation before the user had a chance to choose whether to save. For the broken delete, `DeleteSaveDialogReturned` delegated to `StorageManager.DeleteSaveData()`, whose Windows64 implementation in the prebuilt lib does not delete the folder from disk or invoke the result callback. For text input, the code called `InputManager.RequestKeyboard` on all platforms including Windows64, which resolves to a raw WinAPI `InputBox` with no in-game styling and no controller support.

### New Behavior
After typing a new name and confirming, the save slot updates immediately in the list and persists across restarts. The player's username is correctly displayed on the nametag every time a world is joined, including after saving and re-joining. Choosing "Exit Without Saving" on a brand-new world no longer leaves any folder behind - the world does not appear in the save list at all. Choosing "Delete Save" and confirming removes the save folder from disk and immediately refreshes the list. World naming and renaming now use the existing in-game `UIScene_Keyboard` Flash scene. When a controller is connected the full on-screen keyboard is shown. When only a physical keyboard is available the popup appears with the same visual frame and text field but the on-screen letter and function buttons are hidden, and input is routed directly from the physical keyboard. For the Create World screen on PC without a controller, the world name field supports inline editing in place without any popup.

### Fix Implementation
- **`UIScene_LoadOrJoinMenu.cpp` - `KeyboardCompleteWorldNameCallback`**: Added `#elif defined(_WINDOWS64)` branch. It converts the `ui16Text` input to a wide string and then narrow string, constructs the sidecar path (`Windows64\GameHDD\{UTF8SaveFilename}\worldname.txt`), writes the new name to that file, updates `m_saveDetails[listPos].UTF8SaveName` in memory so the list reflects the change instantly, then calls `RenameSaveDataReturned(pClass, true)` to trigger the existing list repopulate state.
- **`UIScene_LoadOrJoinMenu.cpp` - `ReadLevelNameFromSaveFile`**: Added a sidecar check at the top of the function. Before opening `saveData.ms`, it derives the folder path from the file path, checks for `worldname.txt`, and returns its content immediately if found. This ensures renamed saves display their new name on every load of the save list, not just the session they were renamed in.
- **`UIScene_LoadOrJoinMenu.cpp` - `DeleteSaveDialogReturned`**: Added `#ifdef _WINDOWS64` path that uses `m_saveDetails[displayIdx].UTF8SaveFilename` (the sorted display-order entry) to construct the folder path `Windows64\GameHDD\{name}\`, recursively deletes it using Win32 `FindFirstFileW` / `DeleteFileW` / `RemoveDirectoryW`, and calls `DeleteSaveDataReturned` directly with the success result. The existing console path via `StorageManager.DeleteSaveData` is unchanged.
- **`UIScene_LoadOrJoinMenu.cpp` - `SaveOptionsDialogReturned`**: On Windows64, the rename branch now opens `UIScene_Keyboard` instead of the WinAPI dialog. If a controller is connected (`Win64_IsControllerConnected()`) the keyboard opens in full mode; otherwise it opens in `pcMode` so the on-screen keys are hidden and physical keyboard input is used.
- **`MinecraftServer.cpp`**: Wrapped the automatic `saveToDisc` call for new worlds in `#ifndef _WINDOWS64`. On Windows the first disk write now only happens when the user explicitly saves, so "Exit Without Saving" leaves nothing behind.
- **`Extrax64Stubs.cpp` - `IQNet::EndGame`**: After clearing all player slots, player 0's gamertag is restored from `g_Win64UsernameW` and `m_isHostPlayer` is set back to `true`, so the slot is ready for the next hosted session without requiring a restart.
- **`UIScene_CreateWorldMenu.cpp`**: On Windows64, clicking the world name field without a controller starts inline editing directly in the Flash text field (`m_bDirectEditing = true`). Physical keyboard characters are consumed from `g_KBMInput` each tick. Enter confirms, Escape restores the previous name. A `m_iDirectEditCooldown` counter (4 frames) absorbs the `ACTION_MENU_OK` event that Windows queues alongside `WM_CHAR(0x0D)`, preventing it from re-triggering the edit or the Create World button. When a controller is connected the existing `UIScene_Keyboard` popup is used instead.
- **`UIScene_Keyboard.cpp` / `UIScene_Keyboard.h`**: Added `m_bPCMode` flag. When set, `InitFunctionButtons` is not called and all on-screen keys are hidden: the mapped function-row buttons (`Button_space`, `Button_CursorLeft`, `Button_CursorRight`, `Button_Caps`, `Button_symbols`, `Button_bspace`) via `setVisible(false)`, and the 40 Flash-baked letter/number keys (`Button_q` through `Button_z`, `Button_0` through `Button_9`, punctuation) via direct `IggyValuePathMakeNameRef` + `IggyValueSetBooleanRS` calls identified by decompiling the keyboard SWF. Physical keyboard characters typed in pcMode are accumulated in `m_win64TextBuffer` each tick and pushed to the Flash text field via `setLabel`, bypassing Iggy's focus system.
- **`UIStructs.h`**: Added `UIKeyboardInitData` struct (`title`, `defaultText`, `maxChars`, `callback`, `lpParam`, `pcMode`), `g_Win64KeyboardResult` extern and `Win64_GetKeyboardText` helper, and `Win64_IsControllerConnected()` inline using `XInputGetState` to detect any connected gamepad.

## Related Issues
- Fixes #463 https://github.com/smartcmd/MinecraftConsoles/pull/201#issuecomment-3996507017 #404 #145 #328 
- Related to #463 https://github.com/smartcmd/MinecraftConsoles/pull/201#issuecomment-3996507017 #404 #145 #328

### DIFF
--- a/Minecraft.Client/Common/UI/UIScene_CreateWorldMenu.cpp
+++ b/Minecraft.Client/Common/UI/UIScene_CreateWorldMenu.cpp
@@ -84,6 +84,10 @@ UIScene_CreateWorldMenu::UIScene_CreateWorldMenu(int iPad, void *initData, UILay
 	m_iGameModeId = GameType::SURVIVAL->getId();
 	m_pDLCPack = NULL;
 	m_bRebuildTouchBoxes = false;
+#ifdef _WINDOWS64
+	m_bDirectEditing = false;
+	m_iDirectEditCooldown = 0;
+#endif
 
 	m_bMultiplayerAllowed = ProfileManager.IsSignedInLive( m_iPad ) && ProfileManager.AllowedToPlayMultiplayer(m_iPad);
 	// 4J-PB - read the settings for the online flag. We'll only save this setting if the user changed it.
@@ -289,6 +293,54 @@ void UIScene_CreateWorldMenu::tick()
 {
 	UIScene::tick();
 
+#ifdef _WINDOWS64
+	if (m_iDirectEditCooldown > 0)
+		m_iDirectEditCooldown--;
+
+	if (m_bDirectEditing)
+	{
+		wchar_t ch;
+		bool changed = false;
+		while (g_KBMInput.ConsumeChar(ch))
+		{
+			if (ch == 0x08) // backspace
+			{
+				if (!m_worldName.empty())
+				{
+					m_worldName.pop_back();
+					changed = true;
+				}
+			}
+			else if (ch == 0x0D) // enter - confirm
+			{
+				m_bDirectEditing = false;
+				m_iDirectEditCooldown = 4; // absorb the matching ACTION_MENU_OK that follows
+				m_editWorldName.setLabel(m_worldName.c_str());
+			}
+			else if ((int)m_worldName.length() < 25)
+			{
+				m_worldName += ch;
+				changed = true;
+			}
+		}
+
+		// Escape cancels and restores the original name
+		if (m_bDirectEditing && g_KBMInput.IsKeyPressed(VK_ESCAPE))
+		{
+			m_worldName = m_worldNameBeforeEdit;
+			m_bDirectEditing = false;
+			m_iDirectEditCooldown = 4;
+			m_editWorldName.setLabel(m_worldName.c_str());
+			m_buttonCreateWorld.setEnable(!m_worldName.empty());
+		}
+		else if (changed)
+		{
+			m_editWorldName.setLabel(m_worldName.c_str());
+			m_buttonCreateWorld.setEnable(!m_worldName.empty());
+		}
+	}
+#endif
+
 	if(m_iSetTexturePackDescription >= 0 )
 	{
 		UpdateTexturePackDescription( m_iSetTexturePackDescription );
@@ -354,6 +406,9 @@ int UIScene_CreateWorldMenu::ContinueOffline(void *pParam,int iPad,C4JStorage::E
 void UIScene_CreateWorldMenu::handleInput(int iPad, int key, bool repeat, bool pressed, bool released, bool &handled)
 {
 	if(m_bIgnoreInput) return;
+#ifdef _WINDOWS64
+	if (m_bDirectEditing || m_iDirectEditCooldown > 0) { handled = true; return; }
+#endif
 
 	ui.AnimateKeyPress(m_iPad, key, repeat, pressed, released);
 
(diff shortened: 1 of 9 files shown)


codeHusky:
@Devlrxxh no it does not, I'll merge this one and reopen yours to fix the rest of the menus

PR #583, opened 2026-03-05
Restore broken hurt direction tilt feature from Java Edition

siobhan-saoirse:
## Description
This PR restores functionality of the broken damage tilt feature from Java Edition. There is damage tilt code left over in LCE that is never used anywhere in the client, but only the server applies it with no visual effect.

https://github.com/user-attachments/assets/69f43522-6b1e-4fbb-95b1-0ccd2a5b9d9a


lspepinho:
I suggest adding it as an optional feature in Settings (enabled by default)

BluedragonMask:
Yeah this feels a bit like a view bobbing type thing. I'm with @lspepinho

lspepinho:
> 
> > I suggest adding it as an optional feature in Settings (enabled by default)
> 
> Well unfortunately Idk how to do that
> 

See the FOV slider thing PR

PR #586, opened 2026-03-05
Add Virtual Keyboard to more text inputs

Devlrxxh:
## Description
The current system uses windows DialogBox which looks extremely bad and is unusable on Controller. This introduces a virtual keyboard like on consoles

### Previous Behavior
Changing Texts opens a DialogBox

### Root Cause
Unusable on controller, looks bad

### New Behavior
Opens the new Virtual Keyboard.

### Fix Implementation
There had been issues implementing the callback the issue with the current RequestKeyboard method is that it can't take specific values and I cant change its behavior due to it being built into 4j libs.

### AI Use Disclosure
AI was only used to come up with an idea on how to do the fallbacks as the issue is with the current #RequestKeyboard system its extremely limited.

Preview:
https://streamable.com/shkdzl


void2012:
My point is that there should be no virtual keyboard when an actual keyboard is there and there is no controller. Please implement this behavior.

PR #608, opened 2026-03-05
Removes the Minecraft Store button

FelixPheonix:
## Description
Literally just removes the Minecraft Store button 

## Changes
Changes MainMenu(1080p,720p,480p,Vita).swf to remove the button
Removed the "Minecraft Store" string from the languages.loc file

### Previous Behavior
Minecraft Store button on the main menu did nothing and took up space so I removed it

### Fix Implementation
Used JPEXS Free Flash Decompiler to decompile and change the swf files and removes the button

Minecraft Client_Kv0EfaHmR9


KKNecmi:
🔥

persfidious:
I wouldn't like this for reasons of accuracy. Just make it a stub that does nothing if you really care that much

Dying-Soldier-15:
> I'm unsure if this is needed as that menu could be useful for, example, turning DLCs off/on or managing "Mod" DLCs in the future. This is probably a good change to remove a broken menu in general, but unsure.

I like this idea but I would also like to see the minigames mode either in place of the store button or maybe as another option in the Play Game menu

PR #612, opened 2026-03-05
Overhaul Keyboard/Mouse Support

MrTheShy:
## Description

Full overhaul of KBM (keyboard + mouse) support for Windows64. This started as a follow-up to #539 to fix a few issues with the in-game keyboard, but grew into a proper mouse integration and a complete rework of the direct text editing infrastructure across every UI scene that has text input fields.

## In-game keyboard scene

The native keyboard scene (`UIScene_Keyboard`) syncs the C++ text buffer from Flash at the start of each `tick()` before consuming physical keyboard chars. This keeps both input methods (physical keyboard and on-screen controller buttons) operating on the same text state, so switching between them mid-edit works seamlessly.

The choice between virtual keyboard and direct text input across all scenes is now based on `g_KBMInput.IsKBMActive()` (last active input device) instead of `Win64_IsControllerConnected()` (hardware polling). Using mouse/keyboard gives direct text input, using the controller gives the virtual keyboard, regardless of what hardware is plugged in.

## Mouse hover (UIController.cpp tickInput)

The mouse hover system was rewritten from scratch. The old code used `IggyPlayerGetFocusableObjects` to hit-test against Iggy's internal focusable bounds, which had several structural problems.

Scene lookup is now layer-first across all groups instead of group-first. The old order found tooltip scenes (non-interactive button hint overlays on `eUIGroup_Fullscreen`) before in-game menus on `eUIGroup_Player1`, blocking hover entirely. `eUILayer_Tooltips` is now skipped since those are never interactive.

The hit-test itself now runs against C++ control bounds (`UIControl::UpdateControl` + position/size queries) instead of Iggy focusable objects. Iggy focusable bounds are wider than the visible buttons and overlap vertically, which caused the old "pick largest x0" heuristic to match wrong buttons from the third onward. The new code picks the smallest-area match when bounds overlap and calls `SetFocusToElement` with the already-correct control ID. Accepted types: `eButton`, `eTextInput`, `eCheckBox`, `eSlider`, `eButtonList`.

For `eButtonList` controls, the hover delegates to `SetTouchFocus(mouseX, mouseY, false)` which lets Flash handle internal item highlighting and scrolling. This reuses the Vita touch API — the Flash-side methods were already registered on all platforms in `setupControl()`, only the C++ wrappers were ifdef-gated behind `__PSVITA__`. Changed to `#if defined(__PSVITA__) || defined(_WINDOWS64)`.

Parent panel filtering: each `UIControl` now has a `m_pParentPanel` pointer, set automatically by the `UI_MAP_ELEMENT` macro during `mapElementsAndNames()`. On tabbed menus (LaunchMoreOptionsMenu, CraftingMenu), all controls from all tabs are registered in Flash simultaneously. The hover code checks `ctrl->getParentPanel() != pMainPanel` to skip controls from inactive tabs, preventing phantom hitboxes from overlapping the active tab. `GetMainPanel()` was un-ifdef-ed from `__PSVITA__` for these scenes.

`removeControl()` now calls `setHidden(true)` on all platforms (was Vita-only). The `setHidden`/`getHidden` accessors were un-ifdef-ed from `__PSVITA__`. Both hover and click loops check `getHidden()` to skip removed controls whose C++ objects are still in `m_controls`.

For TextInput controls, `SetFocusToElement` triggers Flash `ChangeState(0)` which shows the yellow highlight border (Iggy's `IggyPlayerSetFocusRS` did not trigger `ChangeState`, so TextInputs had no hover feedback before). `setCaretVisible(false)` is called immediately after to suppress the blinking caret — rendering happens after both tickInput and scene tick, so no flicker.

During direct editing, hover does not move focus to other TextInputs on the same scene (checked via `isDirectEditBlocking()`), preventing e.g. sign line focus from jumping while typing.

When the mouse hovers over empty space, the previous focus target is kept instead of setting `IGGY_FOCUS_NULL`. If a navigation key (arrow) is pressed with nothing focused at all, the first focusable element is grabbed instead of silently dropping the input.

Coordinate conversion uses a single path: window pixel coords to SWF coords via `rawMouse * (sceneRenderDimension / windowDimension)`.

## Mouse click (UIController.cpp + UIScene.cpp)

A virtual `handleMouseClick(x, y)` on `UIScene` hit-tests C++ controls using smallest-area tiebreaker and calls `handlePress`. Returns `true` to consume the click and suppress the `ACTION_MENU_A` dispatch via `m_mouseClickConsumedByScene`. Handles buttons (calls `handlePress`), text inputs (sets Iggy dispatch focus + calls `handlePress`), and checkboxes (toggles state + calls `handleCheckboxToggled`).

For TextInput clicks, a separate `IggyPlayerGetFocusableObjects` pass sets Iggy dispatch focus to the matching focusable, ensuring `ACTION_MENU_OK` targets the correct field.

Click-outside-to-deselect: during direct editing, if the click lands outside the editing text input bounds (adjusted for panel offset), the edit is confirmed via `confirmDirectEdit`. This is handled generically in the base `handleMouseClick` using the `getDirectEditInputs()` list.

Left click maps to `ACTION_MENU_OK`/`ACTION_MENU_A`. Right click maps to `ACTION_MENU_X` (picks up half a stack in inventory/container menus, matching Java Edition). Right click in the world selection menu (`LoadOrJoinMenu`) opens save options (rename/delete), same as RB on controller.

Mouse wheel events (`ACTION_MENU_OTHER_STICK_UP/DOWN`) are centrally remapped to `ACTION_MENU_UP/DOWN` in `handleKeyPress` when KBM is active, so all scenes get scroll support without per-scene handling.

## Crafting menu mouse (UIScene_CraftingMenu.cpp)

Recipe H slots and category tabs are custom-drawn via Iggy callbacks, not Flash buttons, so the standard hover/click system cannot interact with them.

The solution caches SWF-space positions during `customDraw`: when H slot 0 and H slot 1 are drawn, SWF coordinates are extracted from the D3D11 transform matrix via `gdraw_D3D11_CalculateCustomDraw_4J`. The X difference gives uniform slot spacing. Tab hitboxes are derived from the H slot row position since the Vita TouchPanel overlays are full-screen invisible rectangles unsuitable for hit-testing.

`handleMouseClick`: clicking a different recipe slot selects it (updates V slots, highlight, re-shows old slot icon). Clicking the already-selected slot crafts via `handleKeyDown(ACTION_MENU_A)`. Tab clicking runs the same logic as LB/RB. All clicks are consumed to prevent misses from generating `ACTION_MENU_A` and accidentally crafting.

## Direct edit system (UIControl_TextInput)

`UIControl_TextInput` now provides a reusable direct text editing API:

- `beginDirectEdit(charLimit)`: captures current label, inits cursor at end, clears char buffer
- `tickDirectEdit()`: consumes chars from `g_KBMInput`, handles Backspace/Enter/Escape, cursor movement (Left/Right/Home/End/Delete), enforces caret visibility and position every tick (Flash resets both on `setLabel()` and focus transitions), returns Confirmed/Cancelled/Continue
- `confirmDirectEdit()` / `cancelDirectEdit()`: programmatic confirm/cancel
- 4-frame cooldown after confirm/cancel absorbs the `ACTION_MENU_OK` that Enter generates through the input system

`setCaretVisible(bool)` accesses `FJ_TextInput.m_mcCaret` MovieClip to toggle caret visibility. The caret existence is validated once (cached in `m_bHasCaret`/`m_bCaretChecked`) since `IggyValuePathMakeNameRef` always succeeds even for undefined refs. For SWFs without `m_mcCaret` (like AnvilMenu), a `_` character is inserted at the cursor position as a visual fallback, removed on confirm/cancel via `setLabel` with `instant=true`.

`setCaretIndex(int)` calls `FJ_TextInput.SetCaretIndex` to sync the visual cursor position to Flash.

## Direct edit scene integration (UIScene base class)

The base class `UIScene` drives direct editing generically via two virtuals that scenes override:
- `getDirectEditInputs(vector<UIControl_TextInput*>&)`: register text inputs
- `onDirectEditFinished(UIControl_TextInput*, EDirectEditResult)`: handle results

`UIScene::tick()` calls `tickDirectEdit` on all registered inputs and dispatches results. `UIScene::handleMouseClick()` does click-outside-to-deselect. `isDirectEditBlocking()` provides a single guard for `handleInput`/`handlePress`.

**Scenes:**

- **CreateWorldMenu** — world name field. The original ~80 lines of inline push_back/pop_back editing were replaced with `beginDirectEdit(25)` + two override methods.
- **AnvilMenu** — item renaming with direct edit. Live per-keystroke name update: `tick()` compares `getEditBuffer()` against `m_itemName` every frame during editing and calls `updateItemName()` on change, refreshing the repair cost display in real time (like Java Edition). Virtual keyboard opens with `eUILayer_Fullscreen + eUIGroup_Fullscreen` to avoid hiding the anvil container. Null guards on `getMovie()` in `setCostLabel` and `showCross` since the AnvilMenu SWF may not fully load on Win64.
- **SignEntryMenu** — all 4 lines. Auto-starts editing line 1 on first tick if KBM is active. Enter advances to next line, Enter on line 4 confirms the sign. Escape wipes all lines and closes. UP/DOWN arrows move between lines (confirm current, begin edit on new line). DOWN from line 4 moves focus to Done button. The 3D sign shows a blinking `> text <` cursor on the active line via `SetSelectedLine`, driven by `m_iSignCursorFrame`. Done button is clickable during editing (overrides `handleMouseClick` to detect it). `SetSelectedLine(-1)` on destructor clears the 3D cursor.
- **LaunchMoreOptionsMenu** — seed field. Only saves seed on confirm, not cancel.
- **DebugCreateSchematic** — 7 inputs (name + start/end XYZ). Parses ints on finish. All clicks consumed to prevent Iggy re-entry on empty space.
- **DebugSetCamera** — 5 inputs (camera XYZ + Y rotation + elevation). Parses doubles on finish. Float display changed from `%f` to `%.2f`. All clicks consumed.

All scenes support both direct edit (KBM) and virtual keyboard (controller). All keyboard callbacks on Win64 use `Win64_GetKeyboardText(buffer, size)` instead of `InputManager.GetText(buffer)`.

## Other changes

**`UIScene::SetFocusToElement(id)`** — un-ifdef-ed from `__PSVITA__` to also be available on Win64. Used by hover and click code.

**`UIScene::sendInputToMovie` arrow key fallback** — when a navigation key is pressed with no focused element, grabs the first focusable instead of dropping the input.

**Mouse cursor trapped on alt-tab** — `KeyboardMouseInput::Tick()` now checks `m_windowFocused` before calling `SetCursorPos`, so alt-tabbing out actually releases the cursor.

**Escape key during tutorial hints** — removed the `IsTutorialVisible` guard from the KBM pause check. Controller Start button never had this restriction.

**Save buffer crash** — `ConsoleSaveFileOriginal::finalizeWrite()` now checks if the committed buffer is large enough for the full header table before calling `WriteHeader`. New `RegionFile` entries grow the header table without triggering `MoveDataBeyond` (the only place that grows the buffer), causing an access violation on the next autosave when `WriteHeader` tries to write past committed memory. Uses the same `VirtualAlloc` growth pattern already present in `MoveDataBeyond`.

## AI Use Disclosure

Claude 4.6 opus was responsably used as a senior dev in order to:
better understand the codebase
review the code i wrote, 
help me write this PR description in professional English, since English is not my first language. 
Basically all code changes were written by me, the few repetitive and tedious task code that were written with the help of ai was human reviewed and tested locally.

## Related Issues

- Follow-up to #539

### DIFF
--- a/Minecraft.Client/Common/UI/IUIScene_AbstractContainerMenu.cpp
+++ b/Minecraft.Client/Common/UI/IUIScene_AbstractContainerMenu.cpp
@@ -1486,12 +1486,26 @@ bool IUIScene_AbstractContainerMenu::handleKeyDown(int iPad, int iAction, bool b
 		}
 		break;
 	case ACTION_MENU_UP:
+#ifdef _WINDOWS64
+		if (g_KBMInput.WasMouseWheelConsumed())
+		{
+			handleAdditionalKeyPress(ACTION_MENU_OTHER_STICK_UP);
+			break;
+		}
+#endif
 		{
 			//ui.PlayUISFX(eSFX_Focus);
 			m_eCurrTapState = eTapStateUp;
 		}
 		break;
 	case ACTION_MENU_DOWN:
+#ifdef _WINDOWS64
+		if (g_KBMInput.WasMouseWheelConsumed())
+		{
+			handleAdditionalKeyPress(ACTION_MENU_OTHER_STICK_DOWN);
+			break;
+		}
+#endif
 		{
 			//ui.PlayUISFX(eSFX_Focus);
 			m_eCurrTapState = eTapStateDown;
--- a/Minecraft.Client/Common/UI/IUIScene_CreativeMenu.cpp
+++ b/Minecraft.Client/Common/UI/IUIScene_CreativeMenu.cpp
@@ -1099,13 +1099,7 @@ void IUIScene_CreativeMenu::handleAdditionalKeyPress(int iAction)
 		break;
 	case ACTION_MENU_OTHER_STICK_DOWN:
 		{
-			int pageStep = TabSpec::rows;
-#ifdef _WINDOWS64
-			if (g_KBMInput.WasMouseWheelConsumed())
-			{
-				pageStep = 1;
-			}
-#endif
+			int pageStep = 1;
 			m_tabPage[m_curTab] += pageStep;
 			if(m_tabPage[m_curTab] >= specs[m_curTab]->getPageCount())
 			{
@@ -1119,13 +1113,7 @@ void IUIScene_CreativeMenu::handleAdditionalKeyPress(int iAction)
 		break;
 	case ACTION_MENU_OTHER_STICK_UP:
 		{
-			int pageStep = TabSpec::rows;
-#ifdef _WINDOWS64
-			if (g_KBMInput.WasMouseWheelConsumed())
-			{
-				pageStep = 1;
-			}
-#endif
+			int pageStep = 1;
 			m_tabPage[m_curTab] -= pageStep;
 			if(m_tabPage[m_curTab] < 0)
 			{


MrTheShy:
Huge progress in the mouse integration, really, huge, almost 100% working everywhere, also text input is almost perfect everywhere, right now they're not even pushed i am testing a lot, they're wip, will soon be pushed

MrTheShy:
wheel scroll ! xD glad you liked the pr

MrTheShy:
With the controller you can always pause because you have a separate key (start), but we use esc to both simulate b key press and open exit screen, but i've looked better into this issue and yeah, i've now made it so that you can press esc to open the menu always and click b to refuse but this behaviours should be documented somewhere even tho is kinda intuitive

regarding the cannot move mouse it's the same with controller it's intended behaviour

MrTheShy:
Only minor issues in creative mode that i am going to fix later in a new pr, let's merge this, it's already too much stuff for a single pr and it's really a lot of quality of life improvements 

I've pushed, can be merged now in my opinion @smartcmd

MrTheShy:
Fixed both controller and mouse, controlled had  a pre-existent issue where it scrolled 5 pages at a time...

MrTheShy:
@codeHusky can we merge now?

browsertg5:
legend

PR #617, opened 2026-03-05
Xbox360 Media added

Intelinsidecom:
you can now build for Xbox 360 successfully, yes I tested.

## Description
I just added an Xbox media folder copied from Durango with extra file and a header fix so it compiles. I tested on Xenia and it works! credits for XZF file goes to rutinoscorbin on discord. thanks to him for finding how to extract XZF. 
imag33e

Keep in mind I'm not giving exact instructions on how to run this. You need VS 2010 and VS 2012 IDE (Newer proruced bad xex I guess) and Xbox 360 SDK fully installed. also the creator included this py file for copying, very useful
[minecraftbuild2game (1).py](https://github.com/user-attachments/files/25780243/minecraftbuild2game.1.py)

## Changes
Added only one folder, no other changes to other cpp files.

### Previous Behavior
You couldn't build for 360 because Media for them was missing + XZF file.

### AI Use Disclosure
I only asked AI to what the defines should be taken from so I wouldn't define something wrong, causing everything to break, nothing else. it was about the strings.h file in Xbox media folder.


nothingman4ik:
Omg, man nice work

Intelinsidecom:
> One question though: is it true that XBOX SDK maximum supported C++ standard is C++14? Because I (and many others) would like to switch to C++23 and potentially higher. And yes, it _does_ matter.

not sure, probably depends on the xdk version, the newer the better, currently its c++ 11, if it works on modern and old compilers, i think you shouldnt change anything. You can install the 5 gb xdk to test changes if you need. i just added the media folder for people in case they want to build. also switching console version repo for latest c++ is not a good idea since it wouldnt build on the platforms it was designed to be played on. also why you put it as work in progress if i only did one thing, added the folder in case someone wants to build, no file edits were made outside of XboxMedia folder and its everything. nothing else was needed.

void2012:
I assume the problem is the C++ standard library and Microsoft dropped support of that platform at some point?

PR #619, opened 2026-03-05
Fix compiling with cmake

GuglioIsStupid:
## Description
ClientSources.cmake was missing a file, preventing compilation

### Previous Behavior
no compile!!!!

### Root Cause
A file wasn't added to the file list

### New Behavior
do compile!!!

### AI Use Disclosure
who the hell would use ai for one line


GuglioIsStupid:
A script could be made that automatically produces it or smth, idk if GLOB or GLOB_RECURSE would want to be used

PR #629, opened 2026-03-05
Fix mouse getting grabbed outside window of the game

KKNecmi:
## Description

If User uses Windows64 when a ui opens like a furnace does grab(limit your mouse area) your mouse it checks if game in front of the screen if not disable grabbing
https://github.com/user-attachments/assets/70971d59-bcd6-42be-aa04-cfd0227e8112

## Changes
Edited KeyboardMouseInput.cpp & KeyboardMouseInput.h

### Previous Behavior

Alt tabbing causes mouse stuck at center of the screen

### Root Cause

if a ui is open it doesnt check if game is on the screen so it limits everytime effects outside of the game

### New Behavior

if a ui is open it checks if the app window in front of the screen if not dont grab mouse

### Fix Implementation

added a bool to check if app window in front and use this on tick with ifdef _Windows64 check window if its outside game dont grab if its not grab it

### AI Use Disclosure
i used Ai to found the class grabs the mouse so i dont need to look every part of the code
## Related Issues
- Fixes No opened issue
- Related to Nothing

### DIFF
--- a/Minecraft.Client/Common/UI/IUIScene_AbstractContainerMenu.cpp
+++ b/Minecraft.Client/Common/UI/IUIScene_AbstractContainerMenu.cpp
@@ -1486,12 +1486,26 @@ bool IUIScene_AbstractContainerMenu::handleKeyDown(int iPad, int iAction, bool b
 		}
 		break;
 	case ACTION_MENU_UP:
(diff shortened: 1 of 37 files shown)


KKNecmi:
with less file edited this already been done on #612

PR #630, opened 2026-03-05
Project modernization

ModMaker101:
## Description
This replaces all C-style casts in the codebase with their appropriate C++-style equivalent. This improves type safety.

### Previous Behavior
The codebase relied on C-style cases like crazy. The C-style casts lack type safety and can performant multiple types of conversions implicitly.

### Root Cause
The original code used C-style cases because they were new/didn't exist when this was written. This over time can make it harder to program especially when it in C++.

### New Behavior
Most C=style casts now use their C++ equivalent. 

### Fix Implementation
- Went through the codebase and replaced most of the C-style casts with C++ ones.

### AI Use Disclosure
No.

### DIFF
--- a/Minecraft.Client/LocalPlayer.cpp
+++ b/Minecraft.Client/LocalPlayer.cpp
@@ -278,7 +278,7 @@ void LocalPlayer::aiStep()
 	}
 	if (isSneaking()) sprintTriggerTime = 0;
 #ifdef _WINDOWS64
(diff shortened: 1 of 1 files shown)


EnderActually:
> Dang it why is the boat commit there. Oh well.

Should have probably made a separate fork to prevent that.

ModMaker101:
I'll take a look

PR #642, opened 2026-03-06
Fixed the double block break in Creative mode

MarvelcoCode:
## Description
Fixes the double block break in Creative mode when single click

### Previous Behavior
it would break 2 blocks instead of 1

### Root Cause
the tick system Minecraft::tick() never differentiated between single press and a heldpress

### New Behavior
the tick system Minecraft::tick() now differentiates between single press and a heldpress

### Fix Implementation
added differentiations between single press and a heldpress

i would provide a recorded video but it doesnt really show, here is an ss of me playing on my commit 

image


void2012:
Alright, thanks. Is this one ready to merge?

MarvelcoOGFN:
> Alright, thanks. Is this one ready to merge?

yes

PR #649, opened 2026-03-06
fix: ContainerSetSlotPacket and CraftItemPacket item spawning prevention

lag:
## Description

There are numerous vulnerabilities within the packet system right now. This PR clears up some of the major ones.
### Previous Behavior

ContainerSetSlotPacket: let's a player give themselves any item with any nbt.
CraftItemPacket: let's a player craft a recipe without having the required items OR the required crafting table.

### Root Cause

Some comments made it seem like 4J trusted client to compute this. Now that the source is out, it cannot be trusted.

### New Behavior

ContainerSetSlotPacket: blocked completely for survival players. I left it open for creative mode players because some blocks/NBT may not be available through creative menus and may need to be spawned in. Easily changed to a no-op function if that isn't desired.
CraftItemPacket: requires a) crafting recipe be real, b) the person have the required resources, and c) requires the person to have a crafting table open if it's required. (I left 2x2 to not require GUI open, but that can be changed later.)

### Fix Implementation

The new behavior summarizes it well enough I believe, but it's simple gating. If statements that just alert console someone tried to cheat and then returns without computing the cheated items. Nothing crazy.

### AI Use Disclosure

I had an AI agent go through a all of the packets to spot possible vulnerabilities initially, but the actual code written to resolve the issues was handwritten. If that's not an applicable use of AI for this repo, I am sorry.

## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]

### DIFF
--- a/Minecraft.Client/PlayerConnection.cpp
+++ b/Minecraft.Client/PlayerConnection.cpp
@@ -1093,6 +1093,10 @@ void PlayerConnection::handleContainerClose(shared_ptr<ContainerClosePacket> pac
 #ifndef _CONTENT_PACKAGE
 void PlayerConnection::handleContainerSetSlot(shared_ptr<ContainerSetSlotPacket> packet)
 {
+	if(player->gameMode->isSurvival()){ // Still allow creative players to change slots manually with packets(?) -- might want this different.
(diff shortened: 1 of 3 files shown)


codeHusky:
Can you verify that survival crafting / complex recipes still work fine

PR #651, opened 2026-03-06
Fix boat gravity

### LINKED ISSUE #266: Boats fall underground...
Boats on land fall underground... Now you can't iceboat race!

I would've used a video to show this, but Github was failing to import it.
So it's gonna be before, and after.

I placed the boat down.
Image

Then I got into it.
Image

What's weird is it only goes 1 block down.

MozzarellaRat:
Image of boat falling... 
image

## Description & Changes
Fixes #266
Reimplements removed gravity from #615 (#615 unintentionally removes boat gravity)
Removes unreachable code
Comments code
Fixes splash particles rendering anywhere, instead of just in water.

### Previous Behavior
Boats did not have gravity
Boats would fall through land on a passenger being added
Boats would render splash particles when going fast enough, even when not in water.

### Root Cause
#615 improves boat bobbing which also handled gravity.
Boats were intersected with the ground, causing them to fall through the ground
Splash particles did not check for water, just high velocity.

### New Behavior
Boats now have gravity
Boats no longer phase through blocks on ride. 
Splash particles render in water.

## Related Issues
- Fixes #266 
- Related to #615

### DIFF
--- a/Minecraft.World/Boat.cpp
+++ b/Minecraft.World/Boat.cpp
@@ -74,7 +74,7 @@ bool Boat::isPushable()
 Boat::Boat(Level *level, double x, double y, double z) : Entity( level )
 {
 	_init();
-	setPos(x, y + heightOffset, z);
+	setPos(x, y + heightOffset + 0.1, z);
 
 	xd = 0;
 	yd = 0;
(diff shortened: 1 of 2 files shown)


MozzarellaRat:
I do not have a video recorder, sorry. But you can checkout the commit. I could add a photo showing the fix between #266  though

> Could you add a video or smthn? a picture doesn't really show gravity lol

MozzarellaRat:
This also fixes splash particles rendering anywhere, not just water. Forgot to mention that.

PR #669, opened 2026-03-06
Fix crash when FOV is equal to zero

qwasdrizzel:
## Description

Fixes a crash when FOV is equal to zero.

## Changes
Added a if statement inside of `GameRenderer::getFov` to check if `m_fov` is less than 1. If the FOV is less than one it sets it to 0.01.

### Previous Behavior

Crashed when FOV was equal to zero.

### Root Cause

DirectX caused it with the expression `!XMScalarNearEqual(FovAngleY, 0.0f, 0.00001f * 2.0f)`

### New Behavior

Limits FOV to 0.01.

### Fix Implementation

If statement limiting `m_fov` to 0.01 if under 1.

### AI Use Disclosure

No AI.


simulpingus:
when would this crash ever realistically happen

Issue #351, opened 2026-03-04
adjustable render distance


CStone6:
I was messing around in the code, and I saw on line ~611
`renderDistance = (float)(32 * 32 >> (mc->options->viewDistance));`
And if you change this, you get an adjustable render distance. I got ~40 fps with 128 *128, and I got 140 at 5 *5. Also other things I changed might have affected preformace but not that I noticed 

i dont know much about c++ but I think it would be straightforward to link it to buttons or sliders at some point. You would have to look for where the chunks are generated and change the distance because it only renders a small bubble around you, it like voxy where it only shows chunks that were generated. Hope this information is helpful for your guys' endeavors 

<img width="2553" height="1374" alt="Image" src="https://github.com/user-attachments/assets/ce87e729-eae8-43fd-b681-bd8569a945d2" />

<img width="2514" height="1319" alt="Image" src="https://github.com/user-attachments/assets/ee0bfab1-5aea-4620-b280-b8d031e25cdd" />

glisteningtwilights:
i'd love to see a setting like this, this would have a lot of use especially for lower end systems

azloserbits:
I'm gonna mess around and see if I can do it.

Dantez123:
guys i have low fps i wanna change my render distance how to do it i cant find it in the lce windows x64

Dantez123:
> wondering if anyone knows, but I tried to add a slider to the graphics menu, and some config is in a `Loaded iggy movie SettingsGraphicsMenu1080.swf` and it doesn't let you add a menu item without editing, and i dont know how to. Any ideas to get past that?

bro in the game folder how do i change the chunks to 5 or 2 16 is killing my fps i have 3 fps bro

PR #675, opened 2026-03-06
Add Render Distance option.

MijaeLio:
https://github.com/user-attachments/assets/d1bb9fa2-364f-4c23-a78f-819bd7e6cd03

## Description
- Adds a Render Distance option for graphics menu. It can go from 2 to 64 at max.

## Changes
SWF files for SettingsGraphicsMenu on every resolution, also fixing missing FOV option in those files.
GameRenderer.cpp
UIScene_SettingsGraphicsMenu.cpp
UIScene_SettingsGraphicsMenu.h
Consoles_app.cpp
App_enums.h
MediaWindows64.arc

### Previous Behavior
There wasn't a way to change your render distance before. There is mentions for an option called "viewDistance" in GameRenderer.cpp. But it only uses 0 (16 Render Distance) with no way to change it normally.

### Root Cause
No option for this was added on any menu before, and mc->options->viewDistance was unused for anything else than just setting up the game to use 16 chunks render distance only.

### New Behavior
Make mc->options->viewDistance actually useful with adding a slider for it on Graphics Menu

### Fix Implementation
I added two tables for UIScene_SettingsGraphicsMenu.cpp. LevelToDistance and DistanceToLevel to convert between slide values and chunks distances (I actually could do it without using those but it would look ugly for the slider to get 3 - 2 - 1.. values instead of 2 - 4 - 8.. values)

Updated renderDistance on GameRenderer.cpp so the game doesn't glitch when going more than 32 chunks (viewDistance 0)

As said before, updated SWF files to add this option and fix the missing FOV slider for 720 and 480

### AI Use Disclosure
No AI was used. This took me the entire day and was a complete journey even if the changes are small 😭

## Related Issues
- Related to #351


codeHusky (changes requested):
"Temporary" changes should be rolled back before this can be merged

codeHusky:
- delete x64 and restore it back to how it is in Git (discard changes)
- build `main` from this repo
- run the game
- change a setting around, load into a world, save and exit
- compile your pr
- run the game
- open world
- open settings -> graphics
image

the issue is existing settings.dat i guess?  settings.dat from before ur PR results in 0 perhaps

PR #682, opened 2026-03-06
Chat  / Pastes / Formatting

lag:
## Description

This enables Windows 64 native chat from the codebase. Minecraft already had native chat, but it's weirdly disabled on the Windows version. This also adds &n &m &l to the chat since for some reason only colors and &k were supported. (unless I missed something.) It's not a perfect clone, but it's the closest I could get.
### Previous Behavior

Chat would not open and chat messages could not be sent.

### Root Cause

4J didn't enable the chat for Windows 64 version. 

### New Behavior

Uses Iggy-less chat that was in the codebase already and renders properly. It does move the chat down a bit since for some reason they have many different chats. I set the other chat to opacity zero since it uses XUI which didn't like me trying to poke at it's rendering to make colors and stuff work.

### Fix Implementation

Removed blockers that kept the chat locked.

### AI Use Disclosure

Used AI for reviewing my code and to search codebase to find code snippets relating to chat, but everything in commit was written by hand.

### DIFF
--- a/Minecraft.Client/ChatScreen.cpp
+++ b/Minecraft.Client/ChatScreen.cpp
@@ -37,14 +37,14 @@ void ChatScreen::removed()
 void ChatScreen::tick()
 {
 	frame++;
-	if (cursorIndex > (int)message.length())
-		cursorIndex = (int)message.length();
+	if (cursorIndex > static_cast<int>(message.length()))
+		cursorIndex = static_cast<int>(message.length());
 }
 
 void ChatScreen::handlePasteRequest()
 {
 	wstring pasted = Screen::getClipboard();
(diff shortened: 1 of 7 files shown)


lag:
Yeah, I realized I should've recorded something for this when I was making the post... My fault.

https://streamable.com/hl1cgy

It's not perfect, but it's down to just making the bold/italics look better. Gibberish (&k) also pushes characters around, but I'm not sure the best solution for that yet. I'm assuming 99% of players won't even touch color codes until there's a fix for these "problems" though.

Almost all of this comes bundled in with the game. Just takes some opening up of areas in the codebase. The colored text rendering and the chat UI themselves were just in the game files already, but disabled on Windows version.

PR #702, opened 2026-03-06
fix controller breaking when mouse goes off any buttons

Devlrxxh:
## Description
Fixes controller navigation becoming unusable in all menus after moving the mouse off all UI buttons.

## Changes
Previous Behavior
When using Moving your mouse off all UI buttons and then trying to select a button using the controller nothing happens. Basically, the controller is unusable till the mouse is moved back on a button.

## Root Cause
In UIController::tickInput(), the mouse hover logic initializes hitObject to IGGY_FOCUS_NULL and only sets it to a valid handle if the mouse is over a focusable button. When the mouse moves off all buttons, hitObject remains IGGY_FOCUS_NULL, and the code unconditionally calls IggyPlayerSetFocusRS(movie, hitObject, 0) — clearing Iggy's focus to NULL. With no focused element, the controller has no starting point to navigate from.

## New Behavior
When the mouse moves all buttons now the last focused button remains focused, so the controller always has a valid starting button.

## Fix Implementation
Added a hitObject != IGGY_FOCUS_NULL check to the condition before calling IggyPlayerSetFocusRS, so focus is only updated when the mouse is actually hovering over a button.

## AI Use Disclosure
AI was used to identify the issue.

### DIFF
--- a/Minecraft.Client/ChatScreen.cpp
+++ b/Minecraft.Client/ChatScreen.cpp
@@ -1,14 +1,27 @@
 #include "stdafx.h"
 #include "ChatScreen.h"
+#include "ClientConnection.h"
+#include "Font.h"
 #include "MultiplayerLocalPlayer.h"
 #include "..\Minecraft.World\SharedConstants.h"
 #include "..\Minecraft.World\StringHelpers.h"
+#include "..\Minecraft.World\ChatPacket.h"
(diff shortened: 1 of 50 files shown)


Devlrxxh:
This is no longer required the overhaul fixed the issues

PR #748, opened 2026-03-06
Fixed stained glass and glass pane not rendering the water and held item transparency for stained glass and pane

MarvelcoCode:
## Description
Fixed stained glass and glass pane not rendering the water and held item transparency for stained glass and pane

### Previous Behavior
Water wouldnt show through stained glass or glass pane and holding it would look like a solid block

image

### Root Cause
LevelRenderer::renderChunks didnt have a depth mask change for layer 1 (getRenderLayer()) that the stained glass uses and being held rendered as opaque because blending was disabled in  GameRenderer::renderItemInHand

### New Behavior
it now shows the water through stained glass or glass pane and held item isnt opaque anymore

image

image

### Fix Implementation
Added depth mask change for layer 1 (getRenderLayer()) in LevelRenderer::renderChunks and blending was enabled in  GameRenderer::renderItemInHand

### DIFF
--- a/Minecraft.Client/GameRenderer.cpp
+++ b/Minecraft.Client/GameRenderer.cpp
@@ -745,7 +745,12 @@ void GameRenderer::renderItemInHand(float a, int eye)
 			{
 				turnOnLightLayer(a);
 				PIXBeginNamedEvent(0,"Item in hand render");
+				glEnable(GL_BLEND);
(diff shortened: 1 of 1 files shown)


MarvelcoCode:
> @MarvelcoCode Please open a new PR with the fixed changes, this has been reverted

alr will do

PR #767, opened 2026-03-06
Split screen & Overhaul Wide-Screen support 

MrTheShy:
# Description

Split-screen multiplayer for Windows64, plus window resize support, bitmap font fixes, and several crash-causing multiplayer stability fixes. Split-screen was completely non-functional on this platform, window resize wasn't supported, and the networking stack had multiple bugs that surfaced under real multiplayer load (compression buffer overflows, TCP stream desync, chunk visibility races, CompressedTileStorage torn reads, reconnect crashes).

# Changes

### Previous Behavior
Split-screen did not work at all on Windows64. Secondary controllers couldn't sign in (`IsSignedIn` was hardcoded to pad 0), `AddLocalPlayerByUserIndex` was a no-op stub, and `GetXUID` returned invalid for all pads except 0. Even if a second player somehow joined, the 3D world was horizontally stretched because the projection matrix ignored per-viewport aspect ratios. Mouse input only hit the top-left quadrant with no concept of viewport offsets or per-quadrant scaling. The cursor fought between players every tick. On ultrawide monitors the game auto-detected native resolution via `GetSystemMetrics`, breaking the 16:9 assumption in viewport math and Flash layout entirely.

### Root Cause
The Win64 stubs in `Extrax64Stubs.cpp` were never implemented for multi-pad support. `IsSignedIn` only returned true for pad 0, `GetXUID` only returned a valid ID for pad 0, and `AddLocalPlayerByUserIndex` did nothing. `gluPerspective` in `glWrapper.cpp` ignored its aspect parameter and always used the global `g_iAspectRatio` (full-screen 16:9), so split-screen viewports (which are half-width or half-height) got the wrong projection. Mouse coordinate conversion had no concept of viewport offsets or per-quadrant scaling. Screen resolution was pulled from the monitor instead of being pinned to the game's internal 1920x1080.

### New Behavior
Split-screen works on Windows64. Up to 4 players can join with controllers, each getting a correctly rendered viewport with proper aspect ratio. Mouse hover and click work in all quadrants for the primary (KBM) player. Internal resolution is pinned to 1920x1080, keeping viewport math and Flash layout correct on any monitor aspect ratio.

### Fix Implementation
- **Extrax64Stubs.cpp:** `IsSignedIn` now checks `IsPadConnected` instead of hardcoded pad 0. `GetXUID` returns a unique value per pad via `DeriveXuidForPad` (Mix64 hash in `Windows64_Xuid.h`, pad 0 returns base XUID unchanged for save compatibility, includes validity fallbacks if the hash produces an invalid XUID; suggested by rtm516). `AddLocalPlayerByUserIndex` properly initializes the slot with gamertag and remote/host flags. `NotifyPlayerJoined` uses the static `m_player` array directly since `GetLocalPlayerByUserIndex` can't find the player before `customData` is set.
- **glWrapper.cpp:** `gluPerspective` passes the caller's aspect parameter to `RenderManager.MatrixPerspective` instead of the global `g_iAspectRatio`. Split-screen viewports already compute the correct per-viewport aspect in `getFovAndAspect`.
- **UIController.cpp:** Mouse coordinate conversion subtracts viewport origin offset and scales from display size to SWF authoring size per quadrant. Scene lookup restricted to fullscreen group + primary player's group so mouse doesn't bleed into other players' menus.
- **Minecraft.cpp:** Mouse grab/release gated behind `iPad == ProfileManager.GetPrimaryPad()` so split-screen players don't fight over cursor state. Removed `IsTutorialVisible` guard from Escape key pause check.
- **Windows64_Minecraft.cpp:** Removed `GetSystemMetrics` auto-detection, pinned `g_iScreenWidth`/`g_iScreenHeight` to 1920x1080.

---

### Previous Behavior
Only the primary pad on a non-host client had a network connection to the host. Secondary split-screen pads on a non-host machine had no way to send or receive packets, making non-host split-screen multiplayer impossible.

### Root Cause
The networking layer assumed one TCP connection per machine. There was no mechanism for additional local pads to open their own connections, no per-pad socket tracking, and no routing logic to send packets through the correct socket based on which pad originated them.

### New Behavior
Each split-screen pad on a non-host client opens its own TCP connection to the host. From the host's perspective, each connection looks like a normal remote player (gets its own smallId, Socket, PlayerConnection). Secondary pads can join and leave independently.

### Fix Implementation
- **WinsockNetLayer:** `JoinSplitScreen()`, `CloseSplitScreenConnection()`, `SplitScreenRecvThreadProc` for per-pad socket/thread/smallId tracking (`s_splitScreenSocket[]`, `s_splitScreenSmallId[]`, `s_splitScreenRecvThread[]`). `GetLocalSocket()` returns the correct TCP socket for a given local sender's smallId. `GetSplitScreenSmallId()` returns the host-assigned smallId for a pad.
- **GameNetworkManager::CreateSocket:** Non-host path (`localPlayer && !IsHost() && IsInGameplay()`) calls `JoinSplitScreen`, sets the IQNet slot's smallId and resolvedXuid, creates a non-hostLocal Socket + ClientConnection, sends PreLoginPacket, registers via `addPendingLocalConnection`.
- **PlatformNetworkManagerStub::RemoveLocalPlayerByUserIndex:** Implemented the formerly-empty stub. Calls `NotifyPlayerLeaving`, `CloseSplitScreenConnection`, and clears the IQNet slot fields so the pad can rejoin cleanly.
- **SmallId pool:** `s_nextSmallId` starts at `XUSER_MAX_COUNT` (4), reserving `m_player[0-3]` for local pads so remote players never collide.
- **IQNetPlayer::SendData:** Non-host local senders now route through `GetLocalSocket(m_smallId)` instead of always using `SendToSmallId`.
- **IQNet::GetLocalPlayerByUserIndex:** Rewritten. Pad 0 on non-host uses `GetLocalSmallId()` for direct lookup; pads 1-3 check `m_player[padIdx]`.
- **C_4JProfile::IsSignedIn:** Pad 0 always returns true (was checking controller connection, which is unreliable on Win64).
- **GetGamertag/GetDisplayName:** For pads 1-3 with active local players, returns the pad-specific gamertag from `IQNet::m_player` instead of always returning the primary username.
- **ClientConnection:** `isPrimaryConnection()` (true on host or for the primary pad on non-host) guards relative-delta and world-modifying handlers to prevent double-processing of shared state:
  - Guarded: `handleMoveEntity`, `handleMoveEntitySmall`, `handleChunkTilesUpdate`, `handleBlockRegionUpdate`, `handleTileUpdate`, `handleTakeItemEntity`, `handleSignUpdate`, `handleTileEntityData`, `handleTileEvent`, `handleTileDestruction`, `handleComplexItemData`, `handleLevelEvent`, `handleSoundEvent`, `handleParticleEvent`, `handleAddGlobalEntity`.
  - `handleSetEntityMotion`: secondary connections only accept motion targeting their own local player (knockback).
  - `handleExplosion`: world modification (`finalizeExplosion`) guarded, per-player knockback unguarded. Added null check on `localplayers[]`.
  - Entity spawn/remove/teleport/data handlers left unguarded (`putEntity` is idempotent, absolute value setters).
- **handleLogin:** Added else clause to set level when the dimension already exists (was leaving level NULL on reconnect).
- **handleChunkVisibilityArea/handleChunkVisibility:** Added null check on level.
- **handleContainerOpen:** Added null check on `localplayers[m_userIndex]`.

---

### Previous Behavior
Reconnecting to a server after a disconnect would crash or produce undefined behavior. Duplicate XUIDs were rejected with `eDisconnect_Banned`. The old player's write thread could race with the new connection's smallId assignment, sending stale packets on the wrong connection. `SocketInputStream/OutputStream::close()` didn't actually clear their queues. `ServerConnection::stop()` had iterator invalidation bugs and didn't send proper disconnect packets to remote players.

### Root Cause
Multiple race conditions in the disconnect/reconnect flow. The old player wasn't fully cleaned up before the new login was processed. SmallId recycling happened too early (before write threads were dead). Stream close methods called `.empty()` (a read-only check) instead of actually clearing the queue. Server shutdown iterated vectors while modifying them.

### New Behavior
Reconnecting works cleanly. The stale connection is force-disconnected, its smallId is queued for deferred recycling, and the new login proceeds only after the old player is fully removed.

### Fix Implementation
- **PendingConnection:** Duplicate XUID no longer rejects with `eDisconnect_Banned`. Instead it force-disconnects the stale old connection via `stalePlayer->connection->disconnect()`, queues the old smallId for recycling via `queueSmallIdForRecycle()`, then calls `handleAcceptedLogin` for the new connection.
- **MinecraftServer:** Swapped tick order so `players->tick()` (disconnect queue) runs before `connection->tick()` (new logins). The old player is removed from PlayerList before the new LoginPacket's XUID check runs.
- **PlayerList:** `PushFreeSmallId` and `ClearSocketForSmallId` moved here from DoWork, called only after `PlayerConnection::disconnect()` completes and the read/write threads are dead. New `queueSmallIdForRecycle()` method lets PendingConnection push smallIds into `m_smallIdsToClose`, which `PlayerList::tick()` processes through `closePlayerConnectionBySmallId()` for deferred cleanup. Prevents a race where the old write thread could resolve `getPlayer()` to a recycled smallId's new connection and send stale packets on it.
- **SocketInputStreamLocal::close() / SocketOutputStreamLocal::close():** Now actually clear their queues (`std::swap` with empty queue instead of calling `.empty()` which is a read-only no-op).
- **ServerConnection::stop():** Pending and players vectors are snapshot-copied before iterating (prevents iterator invalidation). Remote players receive a DisconnectPacket via `disconnect(eDisconnect_Quitting)` instead of raw `close()`. `tick()`: added else clause so `flush()` only runs on live connections.
- **WinsockNetLayer::Shutdown():** Accept thread stopped first (prevents new recv threads from spawning), then all recv threads are collected and waited on, then connections are closed and split-screen sockets cleaned up. Clears disconnect and free-pool vectors before deleting critical sections.
- **WinsockNetLayer::JoinGame():** Waits for old `s_clientRecvThread` to fully exit before creating a new TCP connection. Prevents the old recv thread from reading bytes off the new socket and desynchronizing the stream.

---

### Previous Behavior
Delayed crashes in unrelated code (`Packet::readPacket`, `LevelRenderer::updateDirtyChunks`) after the first autosave (~30 seconds into a multiplayer game). The crashes appeared random and were extremely difficult to reproduce consistently.

### Root Cause
`CompressLZXRLE` and `CompressRLE` wrote RLE intermediate output into a fixed 100KB buffer with no bounds checking. Full chunk columns are ~160KB and the RLE step can expand 0xFF bytes to 2 bytes each, easily overflowing into `rleDecompressBuf` and heap metadata. The corruption was silent until unrelated code touched the corrupted memory, which is why crashes appeared in `readPacket` or `updateDirtyChunks` rather than in the compression code itself.

### New Behavior
Compression handles arbitrarily large inputs without overflow. Small inputs still use the static buffer (zero allocation overhead). Decompression validates all return values and bounds.

### Fix Implementation
- **CompressLZXRLE / CompressRLE:** Dynamic allocation when worst-case RLE output (`SrcSize * 2`) exceeds the static buffer. Static buffer still used for small inputs. `CompressRLE`: moved `LeaveCriticalSection` after dynamic buffer cleanup.
- **DecompressLZXRLE:** Now checks zlib return value (was completely ignored). On failure, bails out immediately with `*pDestSize = 0`. Added RLE input bounds checking (`pucIn >= pucEnd` before reading count/data bytes) and output bounds checking (`pucOut + count > pucOutEnd`). Same bounds checks applied to `DecompressRLE`.

---

### Previous Behavior
Occasional bad packet ID crashes during multiplayer, especially under load. The crash happened on the client side with an impossible packet ID value.

### Root Cause
The write thread had two output paths to the same TCP socket: `bufferedDos` (5KB buffered stream) and direct `sos->writeWithFlags()`. Chunk data sent via `queueSend()` used the direct path with `shouldDelay=true`, while other packets used `bufferedDos`. If `bufferedDos` had unflushed bytes, the direct write arrived at the client first, reordering the TCP stream and producing bad packet ID crashes.

### New Behavior
All writes to the TCP socket are properly ordered. Direct writes always flush the buffered stream first.

### Fix Implementation
Flush `bufferedDos` immediately before every direct `sos->writeWithFlags()` call.

---

### Previous Behavior
On superflat worlds, one chunk would occasionally be invisible (empty air). On normal worlds with terrain, the same bug crashed the renderer. This happened roughly 30 seconds into a multiplayer session and was tied to the autosave interval.

### Root Cause
`BlockRegionUpdatePacket` (sent via direct socket write through `queueSend`) could arrive at the client before `ChunkVisibilityAreaPacket` (sent through the buffered stream). The client called `getChunk()` on a chunk that didn't exist yet in the cache, got `EmptyLevelChunk` (whose `setBlocksAndData` is a no-op), and silently lost the block data.

A separate race existed in `CompressedTileStorage`: `get()` reads `indicesAndData` twice without a lock. `compress()` can swap the pointer between reads, producing indices from the old buffer paired with data from the new buffer (torn read).

### New Behavior
Block data is never silently dropped. Chunk visibility is set before processing block data, making the handler independent of packet ordering.

### Fix Implementation
- **ClientConnection::handleBlockRegionUpdate:** Calls `dimensionLevel->setChunkVisible()` for full-chunk BRUPs before writing data, making it independent of packet ordering. Added post-write verification logging. Warning log when a full chunk arrives with `ys==0` (empty full chunk, data loss indicator).
- **CompressedTileStorage:** Snapshot `indicesAndData` into a local variable before deriving both pointers. Same snapshot pattern applied to `getData()` (non-Vita path), `isRenderChunkEmpty()`, `getHighestNonEmptyY()`, `getAllocatedSize()`, and `write()`. All methods now also guard against NULL snapshots.

---

### Previous Behavior
The game window could not be resized at all on Windows64. The swap chain was created once at startup and never updated. Resizing the window or switching between windowed/fullscreen either did nothing or crashed.

### Root Cause
Two obstacles prevented standard D3D11 resize:
1. **`ResizeBuffers` fails.** The precompiled `RenderManager` (4J lib) holds internal references to the backbuffer that `Suspend()` does not release. `Suspend()`/`Resume()` are just boolean flag setters (decompiled: `m_bSuspended = true/false`) with no resource release, no thread synchronization, and no memory barriers. Even after releasing every known ref (`g_pRenderTargetView`, the Renderer's own RTV at offset 0x28, its SRV at 0x50, depth stencil at 0x98), calling `ClearState() + Flush()` on the immediate context, `gdraw_D3D11_PreReset()`, and `PostProcesser::Cleanup()`, `ResizeBuffers` still returns `DXGI_ERROR_INVALID_CALL` (0x887A0001). The remaining refs are held by unknown internal code, likely the Renderer's command buffer system, deferred contexts, or TLS-cached render contexts that can't be ClearState'd from outside the lib.
2. **`Initialise()` wipes textures.** `RenderManager.Initialise()` creates all D3D state from scratch (backbuffer RTV, SRV texture, mip-level textures, shaders, index buffers), then does `memset(m_textures, 0, sizeof(m_textures))` — wiping all 512 texture slots (each 40 bytes: `{allocated, ID3D11Texture2D*, ID3D11ShaderResourceView*, flags, mipLevels, format, samplerParams}`) without calling `Release()` on existing textures. The game crashes when scene destruction callbacks fire `TextureFree()` on stale IDs.

The key insight was that `C4JRender RenderManager` is a **stateless wrapper** with zero member variables — all methods delegate to `Renderer InternalRenderManager`. Patching `&RenderManager` directly hit garbage memory. The actual D3D state lives at `&InternalRenderManager`, and its memory layout was mapped by decompiling the 4J lib and cross-checking known pointers at runtime.

Another critical discovery: `StartFrame()` binds `renderTargetView` and `depthStencilView` via `OMSetRenderTargets` and sets the viewport from `backBufferWidth`/`backBufferHeight` **every single frame**. If these members aren't patched to the new dimensions, every frame resets the viewport to the old size, making it look like the resize didn't work even though the swap chain was correctly recreated.

Approaches that were tried and failed: `ResizeBuffers` after Suspend (outstanding backbuffer ref), `ResizeBuffers` with RTV patching (unknown refs remain), direct patching with `&RenderManager` base (no members — offsets hit garbage), force-releasing backbuffer via `Release()` loop (D3D11 debug layer crashes on over-release), swap chain recreation + `Initialise()` (texture wipe), calling `Initialise()` while suspended or before `gdraw_D3D11_PostReset()` (both crash), keeping old swap chain alive alongside new one (crash).

### New Behavior
The game window can be freely resized by dragging borders, maximizing, or switching to fullscreen. The 3D world renders at the new native resolution, Flash UI adapts via Fit16x9 pillarboxing. Rendering is skipped when the window is minimized (avoids 100% GPU usage on a hidden swap chain).

### Fix Implementation
- **Solution: swap chain recreation + direct member patching.** Destroy the old swap chain entirely, create a new one, and directly patch the Renderer's internal member variables via memory offsets. This avoids both `ResizeBuffers` (no outstanding ref issues) and `Initialise()` (no texture wipe).
- **ResizeD3D()** in `Windows64_Minecraft.cpp`: Destroys the old swap chain, creates a new one at the target size via `IDXGIFactory::CreateSwapChain`, then patches `InternalRenderManager` members directly via memory offsets:
  - `0x20` = `m_pSwapChain` (so `Present()` targets the new swap chain)
  - `0x28` = `renderTargetView` (so `StartFrame()` binds the correct render target)
  - `0x50` = `renderTargetShaderResourceView` (so `CaptureThumbnail()` reads from the right SRV)
  - `0x98` = `depthStencilView` (so `StartFrame()` binds the correct depth buffer)
  - `0x5138`/`0x513C` = `backBufferWidth`/`backBufferHeight` (so `StartFrame()` sets the correct viewport)
  - Offsets verified at runtime by cross-checking `m_pDevice` (0x10) and `m_pSwapChain` (0x20) against known globals.
  - Old RTV/SRV intentionally leaked (orphaned with old swap chain) to avoid fighting unknown ref holders.
  - The texture table at offset 0x138 (512 slots, 0x5000 bytes) is left untouched, which is the entire point.
- **Full resize flow:** (1) Guard checks (newW/newH > 0, swap chain exists, `g_bResizeReady`), (2) verify offsets at runtime, (3) `Suspend()` + wait for `Suspended()`, (4) `PostProcesser::Cleanup()`, (5) `ClearState() + Flush()`, (6) release `g_pRenderTargetView`, `g_pDepthStencilView`, `g_pDepthStencilBuffer`, (7) `gdraw_D3D11_PreReset()`, (8) `g_pSwapChain->Release()`, (9) `IDXGIFactory::CreateSwapChain` at new dimensions, (10) patch swap chain pointer at 0x20, (11) `GetBuffer(0)` + `CreateRenderTargetView` + patch RTV at 0x28, (12) create backbuffer-sized texture (`RENDER_TARGET|SHADER_RESOURCE`) + `CreateShaderResourceView` + patch SRV at 0x50, (13) create depth texture + `CreateDepthStencilView` + patch DSV at 0x98, (14) patch `backBufferWidth`/`backBufferHeight` at 0x5138/0x513C, (15) `OMSetRenderTargets` + `RSSetViewports`, (16) `updateRenderTargets()` + `updateScreenSize()` on UIController, (17) `gdraw_D3D11_PostReset()` + `SetRendertargetSize()`, (18) `IggyFlushInstalledFonts()`, (19) `Resume()`, (20) `PostProcesser::Init()`.
- **Critical ordering:** `Suspend()` before any resource release (stops `StartFrame()`/`Present()` calls). `ClearState() + Flush()` before swap chain destruction (unbinds all pipeline state). `gdraw_D3D11_PreReset()` before swap chain destruction (GDraw releases its internal render targets). Swap chain pointer patch immediately after creating the new swap chain (`Present()` must target the new chain when `Resume()` fires). `gdraw_D3D11_PostReset()` before `Resume()` (GDraw must recreate state objects before the next frame). `PostProcesser::Init()` after new swap chain exists (it calls `GetBuffer(0)` to read backbuffer dimensions).
- **WM_SIZE handler:** Defers resize during window drag (`WM_ENTERSIZEMOVE`/`WM_EXITSIZEMOVE`). Immediate resizes (maximize, programmatic) call `ResizeD3D` directly.
- **New globals:** `g_rScreenWidth`/`g_rScreenHeight` (real window dimensions, updated on resize) vs `g_iScreenWidth`/`g_iScreenHeight` (fixed logical resolution, stays 1920x1080).
- **ComputeViewportForPlayer / getFovAndAspect:** Now use `g_rScreenWidth`/`g_rScreenHeight` instead of the fixed startup values, so 3D perspective and split-screen viewports adapt to window size.
- **Main loop:** Rendering skipped when window is minimized (`IsIconic` check).
- **Windows64_UIController:** `ConsoleUIController` caches `m_pRenderTargetView` and `m_pDepthStencilView` as raw pointer copies (NOT AddRef'd). These are passed to `gdraw_D3D11_SetTileOrigin()` every frame during `render()` and `setTileOrigin()` (per-viewport tile origin for split-screen). New `updateRenderTargets(rtv, dsv)` method updates these cached pointers after new views are created during resize.
- **UIController.h:** New inline `updateScreenSize(w, h)` sets `m_fScreenWidth`/`m_fScreenHeight`. These are used by `setupRenderPosition()` (Fit16x9 pillarbox offsets), `getRenderDimensions()` (Flash movie display sizes), and `IggyPlayerSetDisplaySize()` calls during scene rendering. Must only be called when the resize actually succeeds — calling it with new dimensions while the backbuffer is still old-sized causes ghost/double rendering.
- **GDraw (Iggy Flash Renderer) update order:** `gdraw_D3D11_PreReset()` before swap chain destruction (releases GDraw's internal render targets), `gdraw_D3D11_PostReset()` after new swap chain creation (recreates GDraw state objects, must happen before `Resume()`), `gdraw_D3D11_SetRendertargetSize(w, h)` after PostReset (tells GDraw the new backbuffer dimensions). `IggyFlushInstalledFonts()` called after PostReset to flush cached bitmap font textures that were sized for the old resolution.
- **PostProcesser:** Singleton holding offscreen D3D11 resources for gamma correction. `Cleanup()` releases all resources before swap chain destruction, `Init()` called after new swap chain exists — it calls `GetBuffer(0)` to read backbuffer dimensions, then creates new offscreen resources at the new size.
- **InitDevice:** Swap chain `BufferUsage` now includes `DXGI_USAGE_SHADER_INPUT` (needed for the SRV created from the backbuffer for `CaptureThumbnail`).
- **CleanupDevice():** Was leaking `g_pDepthStencilView` and `g_pDepthStencilBuffer`.
- Removed the old `UpdateAspectRatio()` function.

---

### Previous Behavior
At small window sizes, dynamic text (scrollable list items, HowToPlay pages) showed overlapping/inconsistent character sizes. Static SWF text (pre-authored buttons) was unaffected. At intermediate window sizes (e.g. 1678x756), some characters appeared at different scales than others within the same text block.

### Root Cause
Two bugs in `UIBitmapFont.cpp GetGlyphBitmap`:

1. **Small window overlap:** When display scale < bitmap native scale (`pixel_scale < truePixelScale`), `glyphScale` stays at 1, so Iggy displayed the glyph at native 1:1 pixel size but advanced the cursor by the smaller display-scale amount. Result: glyphs wider than their advance, massive overlap.

2. **Intermediate window inconsistency:** Some SWF font sizes produced `pixel_scale` just above `truePixelScale` (13 for Mojangles_11) while others fell just below, splitting glyphs across the small-display and normal cache branches. The normal branch cached all glyphs in a single `[truePixelScale, 99]` range, so the first glyph cached set `pixel_scale_correct` for every subsequent request regardless of font size. Different font sizes then got scaled by wrong ratios (e.g. `18.9/13.3 = 1.42x` with point sampling), producing visibly inconsistent letter sizes.

A third unrelated bug in `UIScene.cpp loadMovie`: the height-based SWF selection could pick 480 or 720 variants on Windows64, which either crashed or loaded the wrong skin library (`skinHD.swf` vs `skin.swf`).

### New Behavior
Dynamic text renders correctly at all window sizes. Characters are consistent regardless of which font size gets cached first. 1080.swf is always loaded on Windows64.

### Fix Implementation
- **UIBitmapFont.cpp:** On `_WINDOWS64`, always use `pixel_scale_correct = truePixelScale` so every cache entry is consistent regardless of which font size creates it first. Two cache ranges: downscale (`pixel_scale < truePixelScale`) uses bilinear for smooth reduction, upscale uses `point_sample` for crisp pixel-art rendering. At most two cache entries per glyph. The console code path (fixed resolution, integer-multiple scaling) is preserved behind `#else`.
- **UIScene.cpp loadMovie:** Always load `1080.swf` on `_WINDOWS64` regardless of window size. Display size is set via `Fit16x9` BEFORE the init tick so Iggy's ActionScript text field creation sees the same scale that `render()` will use. `IggyFlushInstalledFonts()` called after init tick to clear stale glyph cache entries from previous scenes.
- **Font.cpp addCharacterQuad/renderCharacter:** `yOff` was computed with `m_charWidth` instead of `m_charHeight`, producing wrong texture coordinates for non-square glyph cells. This is the world-rendering font (chat, signs, name tags), not the Iggy UI font.

---

### Previous Behavior
The in-game keyboard scene had no cursor movement support in PC mode. Text could only be appended or deleted from the end.

### New Behavior
Full cursor navigation in PC mode: arrow keys move the cursor, Home/End jump to start/end, Delete removes the character after the cursor. Flash caret syncs with cursor position.

### Fix Implementation
- **UIScene_Keyboard.cpp:** Added `m_iCursorPos` tracking for PC mode with full arrow key, Home, End, Delete support and Flash caret sync via `SetCaretIndex`.

---

## 10. Misc fixes and diagnostics

- **Packet::readPacket:** Thread-local ring buffer tracks last 8 good packet IDs. On bad packet ID, dumps the history plus next 32 bytes of stream for diagnosing TCP desynchronization.
- **Textures::releaseTexture:** Early return for `id <= 0`, checks `TextureGetTexture(id) != NULL` before calling `glDeleteTextures`. Prevents crashes on stale texture IDs after RenderManager reset.
- **UIController TextureSubstitutionDestroyCallback:** Null guard on `Minecraft::GetInstance()` and `mc->textures` before calling `releaseTexture`. Prevents crash during shutdown.
- **StringTable:** Removed `__debugbreak()` on language load failure in debug builds.
- **PendingConnection/PlayerList:** Debug logging for the reconnect flow (duplicate XUID handling, force-disconnect, `handleAcceptedLogin`, `placeNewPlayer` with smallId/entityId/dimension).

---

## Known issues / future work

**SendOnSocket global lock** (`WinsockNetLayer.cpp`): `s_sendLock` is a single CriticalSection serializing ALL TCP sends across ALL connections. If one client's `send()` blocks (TCP window full, slow network), every other write thread stalls. Each PlayerConnection has its own write thread, so with 8+ players one slow client can cause latency spikes or timeout disconnects for healthy players. Fix: replace `s_sendLock` with per-socket locks indexed by smallId. Deferred to a separate PR.

# AI Use Disclosure
Claude 4.6 Opus was responsibly used as a senior dev assistant throughout this work:
- **Codebase understanding:** Navigating unfamiliar parts of the codebase (4J lib internals, Iggy/Flash rendering pipeline, D3D11 resource lifecycle), understanding call chains and data flow through decompiled 4jlibs code.
- **Code review:** Reviewing the code I wrote for correctness, spotting edge cases and potential race conditions.
- **Debugging:** Brainstorming possible root causes when facing obscure crashes, helping trace through execution paths to narrow down bugs (e.g. the compression buffer overflow, the TCP stream desync, the chunk visibility race).
- **Architecture exploration:** Helping map out the InternalRenderManager memory layout, understanding why `ResizeBuffers` and `Initialise()` couldn't work, suggesting which offsets to investigate and why.
- **PR description:** Helping write this description in professional English, since it's not my first language.

All code changes were written by me. The few repetitive and tedious pieces of code that were written with AI assistance were human-reviewed and tested locally. All architecture decisions, debugging strategies, and implementation choices are mine.


MrTheShy:
yes, i will commit, then merge main into this then remove draft

MrTheShy:
Took some time but it's now perfect

codeHusky:
Thanks for this, awesome PR! fixes a ton of bugs

PR #816, opened 2026-03-07
Add IPv6/Dual Stack

12brendon34:
## Description

Modifies WinsockNetLayer.cpp to use Dual stack/IPv6 support

## Changes
Add a little helper WinsockNetLayer::IsNumericAddress,
Modify HostGame, AcceptThreadProc

### Previous Behavior

IPv6 support just didn't exist

### Root Cause

It's old

### New Behavior

It now uses the IPv6 flags and IP parsing logic

### Fix Implementation

```cpp
bool WinsockNetLayer::HostGame(int port, const char* bindIp)
{
...
	hints.ai_family = AF_UNSPEC;
...
	if (resolvedBindIp == NULL)
		hints.ai_flags = AI_PASSIVE;
	else if (IsNumericAddress(resolvedBindIp))
		hints.ai_flags = AI_NUMERICHOST;
	else
		hints.ai_flags = 0;
...
	DWORD ipv6only = 0;
	setsockopt(s_listenSocket, IPPROTO_IPV6, IPV6_V6ONLY, (const char*)&ipv6only, sizeof(ipv6only));
...
}

bool WinsockNetLayer::JoinGame(const char* ip, int port)
{
...
	if (IsNumericAddress(ip))
		hints.ai_flags = AI_NUMERICHOST;
	else
		hints.ai_flags = 0;

	hints.ai_family = AF_UNSPEC;
...
}


bool WinsockNetLayer::IsNumericAddress(const char* addr)
{
	if (addr == NULL)
		return false;

	if (strchr(addr, ':') != NULL)
		return true;// IPv6

	if (inet_addr(addr) != INADDR_NONE)
		return true;// IPv4

	return false;// hostname
}
```

### AI Use Disclosure

No AI was used for this


12brendon34:
I've tested it with one of my friends, it does seem to work quite well.
I'll fix the conflicts tmr and if anyone else wants to test it, feel free to.
I'll likely remove the servers.dat code and make a separate pull request, probably should have done so from the beginning.

12brendon34:
I think this is suitable to be merged now.
I've done all the testing I can on my setup.

ApfelTeeSaft:
> > unrelated to the implementation, but when this is merged into the main branch i can just set a AAAA record to an ipv6 and enter the assigned hostname in the server list right?
> 
> Yea, that's what I use.
> image
> works fine over starlink and such.

But is it already implemented, the latest version for me doesn't seem to work over ipv6, either I'm an idiot with setting up my domain, tho I did give access to my ipv6 on my router settings and it works fine if I run a java server, so I'm confused asf

PR #880, opened 2026-03-07
Add option to control mob cap implementation

Alexandra-Myers:
## Description
This is more or less an expansion of #789, preserving LCE functionality by default whilst letting the user remove the mob cap.
This PR introduces per-chunk caps based on Java Edition 1.8.8, as well as 4J's comments on what they changed from Java.

### Previous Behavior
Previously, the limits for certain mob groups in their count was unchangeable, and the limits were enforced level wide rather than per chunk.

### Root Cause
This was not erroneous but was rather a design decision to allow for better performance on the devices it ran on.

### New Behavior
By default, it uses the same logic as existed before this version, however there is now an option to turn off mob caps, when true it is as follows:
Does NOT remove the mob cap for spawning, rather, breeding, spawn eggs, and construction are all unlimited, and spawning works exactly like in Java, with the base size per chunk for the mob category (leftover in original LCE code, though I edited it to keep the max count the same when one player is on) being scaled by the amount of loaded chunks and then scaled back down by a magic number (it is literally called MAGIC_NUMBER in official mappings for the modern game), which is then used as the mob cap for that category, effectively acting as the mob caps used in LCE, except more dynamic.

### Fix Implementation
Stores `eGameHostOption_WorldMobCap` onto game host options using the bitmask 0x40000000.
Added `no-mob-cap` to server.properties to control it on the dedicated server.
Added a checkbox to `UIScene_LaunchMoreOptionsMenu` for modifying the mob cap setting.
Inside of `Level::canCreateMore`, added a check for if `eGameHostOption_NoMobCap` is enabled, returning true early if so.
Removed `MobCategory::getMaxInstancesPerLevel`, its usage has been replaced with `MobCategory::getMaxInstancesPerChunk`.
`m_max` in `MobCategory` was turned into a pointer as a reference for the now non-constant caps. (later can become user controlled with a better config system, ideally)
Removed the `m_maxPerLevel` field from `MobCategory`, instead using `m_max`.
Inside of `MobSpawner::tick`, references to `mobCategory->getMaxInstancesPerLevel()` were moved to their own variable and replaced with `mobCategory->getMaxInstancesPerChunk()`, as now if `eGameHostOption_NoMobCap` is enabled, the max instances will be updated to instead use Java's math, as so:
```cpp
maxInstances = mobCategory->getMaxInstancesPerChunk() * chunksToPoll.size() / MAGIC_NUMBER; // Java is mobCategory.getMaxInstancesPerChunk() * this.spawnableChunkCount / NaturalSpawner.MAGIC_NUMBER
```

### AI Use Disclosure
No AI was used in the making of this Pull Request.

## Related Issues
- Related to #789


Alexandra-Myers:
> does anyone have any suggested software for dealing with swf files?

I know what they are, however I don't know how to use them.

Alexandra-Myers:
> Please check your merge conflicts!

I did not really have the time before, I have fixed the merge conflicts.

PR #886, opened 2026-03-07
fix: potential rces, reimplement java style exceptions, implement sane checking in netcode (#794)

### LINKED ISSUE #794: [Bug, Security] potential security problems with Packet class, asserts in general
### Bug description

`Minecraft.World/Packet.cpp` acts as the backbone for most of the multiplayer systems inside of MCLCE, however i've noticed that most of the checks done (e.g `readUtf` just passes through any string, it throws an assert but those are usually no-opped on release builds). 

there's an absurd amount of developer comments left from 4j basically warning of this
```cpp
// throws IOException TODO 4J JEV, should this declare a throws?
```
most of the packet code is cloned from the code from java edition, and all the proper error handling seems to be implemented but only in java, an `assert` was used as a placeholder until they (4j) actually figured out a way to handle said errors. 

### To Reproduce

an example of this can be seen in `Packet::readUtf`

https://github.com/smartcmd/MinecraftConsoles/blob/4f73dd18b109b29abba2102a65fcf36fc8a32298/Minecraft.World/Packet.cpp#L393-L418

the first check is done to see if the string is longer than the max string length (in a signs case, 15 characters) if it is, it writes a short message along with the `stringLength` and the `maxLength` to a variable called `stream`, as a far as i can tell, nothing is done with this.

there's an assert used here, only a debug build (and the current way release builds are configured) notice this, and stop the game (at least until you tell it to ignore the assert).

the other check is pretty much the same story, just an assert that doesn't really go anywhere.

### Expected behavior

N/A

### Screenshots

wine debugger assert error<br>
an example of an assert thrown by a manually passed (via python) invalid string of an invalid length.

### Version

16eec7e

### Build Type

- [x] Release
- [x] Debug

izzint:
fixes #794

**This PR is not fully finished**, going through netcode takes a lot of effort and there's more and more new stuff i keep finding, expect a full write-up here once this PR gets to a more complete and all-encompassing state.

Currently implements `IOException`, `EOFException`, `RuntimeException`, and `IllegalArgumentException`.
trying my best to fill in the voids left by 4j where these are supposed to fit, as well as finding new places to insert them. 

For more details on the current state of this, please check the commit history.

Please keep in note that contributions from other people are extremely welcomed! ;]

### DIFF
--- a/Minecraft.Client/ClientConnection.cpp
+++ b/Minecraft.Client/ClientConnection.cpp
@@ -142,6 +142,13 @@ ClientConnection::ClientConnection(Minecraft *minecraft, Socket *socket, int iUs
 	deferredEntityLinkPackets = vector<DeferredEntityLinkPacket>();
 }
 
+bool ClientConnection::isPrimaryConnection() const
+{
+	// On host, all connections are primary (server is authoritative).
+	// On non-host, only the primary pad processes shared entity state.
+	return g_NetworkManager.IsHost() || m_userIndex == ProfileManager.GetPrimaryPad();
+}
+
 ClientConnection::~ClientConnection()
 {
 	delete connection;
@@ -328,6 +335,10 @@ void ClientConnection::handleLogin(shared_ptr<LoginPacket> packet)
 			level->difficulty = packet->difficulty; // 4J Added
 			level->isClientSide = true;
 			minecraft->setLevel(level);
+		}
+		else
+		{
+			level = (MultiPlayerLevel *)dimensionLevel;
 		}
 
 		minecraft->player->setPlayerIndex( packet->m_playerIndex );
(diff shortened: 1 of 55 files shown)


izzint:
> If so, try finding a solution that doesn't involve exceptions, if possible.

will do, thanks for the feedback.

codeHusky:
how close is this?

izzint:
> how close is this?

i haven't had too much time today to work on it, i'm going to be busy with school tomorrow but i'll probably get off around noon my time, i could probably finish it then.

codeHusky:
#1023 may have addressed some things in this and has since been merged, you may need to do some conflict resolution. your method of disconnecting people is superior to the fail silently fixes in that PR though so feel free to swap out or add disconnects as needed

Issue #864, opened 2026-03-07
[Suggestion] Old Swinging Animation Toggle


Terralys:
### Describe the solution you'd like

Hey, I was wondering if it is a possible to add the old swinging animations back to the release build of the project but as an option? I know you guys fixed the buggy animation because I saw it in the github but I was wondering if we could get a option to turn on the old swinging animations? I know this might seem stupid but the main reason I don't play legacy4j and those mods is because they don't have the swing animation. If you could please add this in it would be great, I really like what you guys have already done here and the only thing to me that feels missing is an option to toggle which swinging animation you prefer. 

### Is your suggestion related to a problem?

A Bug that was fixed.

### Additional context

I know this is a very stupid request but I'd really appreciate it and I'm sure those who have nostalgia and a liking for the old buggy animations might also appreciate this. The op that suggested that fix also had this to say "i think that there should be a toggle for this fix for the future" this bug was fixed here, Please add an option to bring these animations back. https://github.com/smartcmd/MinecraftConsoles/pull/418

sudo-lcy:
i second this, with the 'fixed' swing it just felt like i'm playing any other minecraft flavours with a legacy console UI slapped on top of it, it would be nice to have a legacy swing toggle

PR #894, opened 2026-03-08
Fancy Graphics, Old Swing & Smooth Lighting checkboxes in options.

MijaeLio:
https://github.com/user-attachments/assets/9f781882-b389-43c0-86da-b2703259dc5c

https://github.com/user-attachments/assets/b0b5ff47-248f-4494-ab85-19638ef939e9

## Description
Introduces checkboxes for Ambient Occlusion & Fancy Graphics, for enabling and disabling them.

## Changes
- SettingsGraphicsMenu swf files.
- App_Defines.h
- App_enums.h
- UIScene_SettingsGraphicsMenu both .cpp and .h
- Consoles_app.cpp
- Extrax64Stubs
- GameRenderer
- TileRenderer
- MediaWindows64.arc

### Previous Behavior
Ambient Occlusion and Fancy Graphics options were unused, left in the code but working.

### Root Cause
There wasn't any option in-game to switch them.

### New Behavior
You get new checkboxes for Graphics options for AO and FG. 

### Fix Implementation
Edited SWF files to add the checkboxes, Consoles_app.cpp so the settings save properly and for having effect on those options in-game, SettingsGraphicsMenu to add the new checkboxes in the code, defined GAMESETTING for the new options using uiBitmaskValues, restored function for AO in TileRenderer so it matches with the option.

### AI Use Disclosure
No AI was used.

## Related Issues
- Related to #864

### DIFF
--- a/Minecraft.Client/ClientConnection.cpp
+++ b/Minecraft.Client/ClientConnection.cpp
@@ -142,6 +142,13 @@ ClientConnection::ClientConnection(Minecraft *minecraft, Socket *socket, int iUs
 	deferredEntityLinkPackets = vector<DeferredEntityLinkPacket>();
 }
 
+bool ClientConnection::isPrimaryConnection() const
+{
+	// On host, all connections are primary (server is authoritative).
+	// On non-host, only the primary pad processes shared entity state.
+	return g_NetworkManager.IsHost() || m_userIndex == ProfileManager.GetPrimaryPad();
+}
+
 ClientConnection::~ClientConnection()
 {
 	delete connection;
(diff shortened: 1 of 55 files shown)


codeHusky:
Ambient Occlusion should probably be renamed to “Smooth Lighting” to be more consistent with how it’s explained in Java edition, unless there’s some precedent for calling it that in LCE

MijaeLio:
> Since the other person mentioned how it makes it look more like TU1, maybe you could batch in a toggle for #864 here since it's also a very similar change? It was changed in #418 , hopefully it'd be simple to change it to be controllable with a checkbox: https://github.com/smartcmd/MinecraftConsoles/pull/418/changes
> 
> I was going to do it myself, but since I would have been doing this alongside you it would have conflicted with your changes and need manual merging, so I think it would be better for maintainers if done here.

Just added this in! On user interface options tho, but it is in

codeHusky:
Other issue, graphics settings are not persisting across relaunches

PR #911, opened 2026-03-08
Add "Add Server" functionally to "Join Game" Menu + relocate servers.txt to servers.db. 

mattsumi:
## Description
replaces the old plaintext servers.txt format with a structured binary servers.db file and adds an in-game "add server" using the native ui so users no longer need to manually edit a text file.

### Previous Behavior
servers were loaded from a plaintext servers.txt file that users had to manually create and edit with a specific line-based format (ip, port, name repeating). there was no way to add servers from within the game.

### Root Cause
the txt-based approach was a temporary solution that required users to alt-tab out, edit a file by hand, and restart. it was error-prone and not user-friendly.

### New Behavior
servers are now stored in a binary servers.db file using a versioned format (magic "MCSV", (literally just MineCraft Servers lmfao) version 1) with length-prefixed ip, port, and name fields. a multi-step keyboard input flow (ip -> port -> name). from this you can just join the server. as this is cached, you no longer need to wait for it to populate the server area.

### Fix Implementation
[PlatformNetworkManagerStub::SearchForGames()] now reads servers.db (binary) instead of servers.txt (plaintext) and parses the versioned binary format alongside lan discovery results.

for the native ui changes, [UIScene_LoadOrJoinMenu] gains an [eAddServerPhase] state machine (peAddServer_Idle], [eAddServer_IP], [eAddServer_Port], [eAddServer_Name]) and members [m_addServerIP]/[m_addServerPort] to drive the three-step keyboard input as mentioned previously.

### AI Use Disclosure
No AI was used for this PR.


lspepinho:
This is pretty good imo, but you should resolve merge conflicts

codeHusky:
Does this support deleting or renaming servers?

PR #921, opened 2026-03-08
fix: prevent TNT minecarts from exploding if TNT host option is disabled

### LINKED ISSUE #904: [Bug] Disabling TNT does not disable TNT minecart explosions
### Bug description

Disabling TNT does not disable TNT minecart explosions

### To Reproduce

go on a server with TNT disabled, and spawn a TNT minecart and let it drop from a height/get set on fire.

### Expected behavior

it should vanish or do nothing.

### Screenshots

_No response_

### Videos

_No response_

### Version

a865cea

### Build Type

- [x] Release
- [x] Debug

### Additional context

Please make this urgent. This means griefing and server crashing is very easy (see my other issue about TNT cart server crashers.)

Alezito2008:
## Description
Fixed a gameplay inconsistency where TNT Minecarts would ignore the "TNT Explodes" host option, allowing them to explode even when TNT was disabled for the world.

### Previous Behavior
TNT Minecarts would always explode regardless of the world's host settings. This allowed for accidental or intentional griefing in worlds where the host specifically disabled TNT.

### Root Cause
The `MinecartTNT::explode` function lacked a check for the `eGameHostOption_TNT` flag

### New Behavior
TNT Minecarts now respect the TNT explodes option.

### Fix Implementation
Wrapped the explosion and entity removal logic inside a conditional check for `app.GetGameHostOption(eGameHostOption_TNT)`

### AI Use Disclosure
No

## Related Issues
- Fixes #904

### DIFF
--- a/Minecraft.World/MinecartTNT.cpp
+++ b/Minecraft.World/MinecartTNT.cpp
@@ -80,9 +80,12 @@ void MinecartTNT::explode(double speedSqr)
 	if (!level->isClientSide)
 	{
 		double speed = sqrt(speedSqr);
-		if (speed > 5) speed = 5;
-		level->explode(shared_from_this(), x, y, z, (float) (4 + random->nextDouble() * 1.5f * speed), true);
-		remove();
+		if (speed > 5.0) speed = 5.0;
+		if (app.GetGameHostOption(eGameHostOption_TNT))
+		{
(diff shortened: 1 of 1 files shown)


Alezito2008:
https://github.com/user-attachments/assets/7f5de94e-58d2-468f-a65e-3eb42294f093

eh-K:
Nice find 🧠

Issue #954, opened 2026-03-08
[Suggestion] Discord rich presence


nitrate92:
### Describe the solution you'd like

Discord rich presence that would make use of the old rich presence strings you would've seen on PSN or Xbox Live

i.e. "Enjoying the view!" or "In the menus", they are in some of the strings files under `IDS_RICHPRESENCESTATE` and if you remember them you'll already know what I'm talking about

xgui4:
imo i will like to be able for this feature to be disable at compile time and on setting, because i am not the greatest fan of discord and i do not trust them.

PR #968, opened 2026-03-08
Adding discord rich presence support for Windows64 (Early)

acth2:
## Description
Reintroducing the Discord rich presence system, this idea come from #954

### Previous Behavior
There were the functions for rich presence but no code inside of them: the code is still in an very early state but it works!
image

### Root Cause
I needed to recode the discord rich presence but the system that tracks the players action and that permit me to modify what the rich presence says was already coded!

### New Behavior
in Minecraft:Init i call ProfileManager.RichPresenceInit(...) and in Minecraft::tick i call player->updateRichPresence(); to constantly update the rich presence
NEW: I now use the Minecraft.Client/Windows64Media/loc/stringsRichPresence.xml file that contains the rich presence text (thanks to eff3ry for pointing that)

### Fix Implementation
Previously it looked like this:
`void				C_4JProfile::RichPresenceInit(int iPresenceCount, int iContextCount) {}`
`void				C_4JProfile::SetRichPresenceContextValue(int iPad, int iContextID, int iVal) {}`
i just called discord-rpc.h and followed instructions to make a rich presence from there:
image

_It works but its ugly for now!_

### AI Use Disclosure
nop

## Related Issues
- Related to #954


eff3ry:
would it be better to hook directly into the localised strings?
`Minecraft.Client/Windows64Media/loc/stringsRichPresence.xml`

SimplyCEO:
Could this code be placed under a definition, please?
I do not have Discord, so, how I see, this is extra unused binary.

Issue #838, opened 2026-03-07
[Bug] FOV 110 causes everything to FLIP when sprinting?


YeetTeamYT:
### Bug description

Video [`here`](https://youtu.be/7YgBqreTwmg)

### To Reproduce

1. Download latest nigltly
2. Set FOV to 110
3. Open/create a world
4. Try sprinting

### Expected behavior

I expected the game to be normal, not flipping everything when i sprint at FOV 110

### Version

Nigltly build released 45 Mins ago at time of post

### Build Type

- [x] Release
- [ ] Debug

Issue #802, opened 2026-03-07
[Bug] FOV and Render Distance


UnLimitedCaNNoNs:
### Bug description

the fov slider in tonights update is very off, i find it has become too exaggerated: especially whilst sprinting. and with the addition of the render distance slider, i find it very useful but it has brought along some fps drops. ive tested with last nights build to compare and i find that even with a default render distance of 16, the fps drops very noticeably on tonights build: i have to set it to a distance of 8 to get the smoothness of last update 

### To Reproduce

as stated above, the render distance will cause a lowering in fps if above 8 chunks, in comparison to the last updates default of 16 chunks without any drop in fps

along side this, like i said beforehand, the fov has become too exaggerated: especially: whilst sprinting. idk what cuases this 

### Expected behavior

makes shit slower than noraml and the camera looks weird at higher fps

### Version

16eec7e

### Build Type

- [ ] Release
- [x] Debug

### Additional context

i should say, that the fov slider in last nights update was way less sensitive in regard to the update for tn

Issue #771, opened 2026-03-06
[Bug] FOV slider is not clamped to Java Edition values


Eribetra:
### Bug description

The FOV slider does not match Java Edition's in higher settings, and enables you to scale your FOV to values much higher than what is supposed to be FOV 110/Quake Pro. This results in an unusually, unusably high visual FOV in the maximum setting, alongside a flipped view when running while flying.

### To Reproduce

1. Change the FOV slider to a value above 88

### Expected behavior

The resulting visual FOV matches Java Edition's; FOV 110 in LCE should look the same as FOV 110 in Java Edition.

### Screenshots

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/8899601d-a824-4a19-874d-a3c51d14f647" />

### Videos

https://github.com/user-attachments/assets/bf924d97-7b0a-4132-9ec1-b7036fcb68

https://github.com/user-attachments/assets/f1352301-a1cf-459e-a7c7-89bdecfc212c

73

### Version

 f254af3

### Build Type

- [x] Release
- [ ] Debug

### Additional context

From eye-ball testing, it seems like FOV value 88 in LCE matches 110/Quake Pro in Java Edition.

Issue #666, opened 2026-03-06
[Suggestion] Save user-defined FOV value


Eribetra:
### Describe the solution you'd like

The FOV value should be saved so it can persist on relaunches.

### Is your suggestion related to a problem?

The FOV value is currently not saved in between reloads of the game. When restarting, it is reset to 70, regardless of what value the user has set before.

### Additional context

https://github.com/user-attachments/assets/88eb47e5-3beb-46d9-8946-2e1344a43955

PR #975, opened 2026-03-08
Fix FOV Slider + Viewmodel

### LINKED ISSUE #655: [Bug] Hand FOV changes when the FOV is changed in graphics settings
### Bug description

https://github.com/user-attachments/assets/3b9e2690-8035-4d18-a33e-8a88cacddf9f

### To Reproduce

1. Change fov

### Expected behavior

Hand should stay at the same FOV regardless of the FOV setting, as it does in every other edition.

### Screenshots

_No response_

### Videos

_No response_

### Version

0c2e27c

### Build Type

- [x] Release
- [ ] Debug

### Additional context

_No response_

jvnpr:
# [!] See #1135

## Description
This PR fixes the FOV slider to use the pre-existing FOV option as intended instead of hooking into gameRenderer. It also fixes some other minor issues with the existing FOV system and increases the range of FOV to 30-110.

### Previous Behavior
The FOV slider was inaccurate because it modified a value gameRenderer expected to be hardcoded.

java110
lcenightlybuild110

Additional Issues: 
- FOV slider is not 1:1 with FOV values
- Hand viewmodel is affected by FOV changes

### Root Cause
The FOV slider implemented by 4J in the debug menu (which was used as a reference for the slider in settings) modified the m_fov value in gameRenderer instead of modifying the existing option. 

The FOV slider was also implemented as a 0-100 slider and the value was used to calculate an FOV, meaning that multiple stops of the slider corresponded with a single FOV value

The hand renderering code calls for the target FOV, meaning it is affected by FOV scaling. 

### New Behavior
The FOV slider now modifies the intended option value that gameRenderer expects it to. The FOV slider also now goes from 30-110 instead of 70-110. 

java110
fovtweak110

java30
fovtweak30

The FOV slider is now 1:1, with each tick / stop corresponding to one discrete FOV value. 

The hand viewmodel no longer scales with screen FOV.

### Fix Implementation
Instead of using `gameRenderer->SetFovVal();`, the slider now modifies `options->fov`. This change is used in UIScene_SettingsGraphicsMenu.cpp, UIScene_DebugOverlay.cpp, and Consoles_App.cpp.

Changed how `m_sliderFOV` was implemented, making it 0-80 to match the range of 30-110

Added a hardcoded value into `gluPerspective();` in `GameRenderer::renderItemInHand();`

### AI Use Disclosure
No AI was used to modify the code or author this pull request. 

## Related Issues
- Fixes #655 #666 #771 #838
- Related to #802

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,0 @@
-*.png filter=lfs diff=lfs merge=lfs -text
-*.jpg filter=lfs diff=lfs merge=lfs -text
-*.ogg filter=lfs diff=lfs merge=lfs -text
-*.binka filter=lfs diff=lfs merge=lfs -text
-*.arc filter=lfs diff=lfs merge=lfs -text
-*.ttf filter=lfs diff=lfs merge=lfs -text
-*.bin filter=lfs diff=lfs merge=lfs -text
-*.ico filter=lfs diff=lfs merge=lfs -text
--- a/Minecraft.Client/Chunk.cpp
+++ b/Minecraft.Client/Chunk.cpp
@@ -739,9 +739,9 @@ void Chunk::rebuild_SPU()
 				{
 					// 4J - get tile from those copied into our local array in earlier optimisation
 					unsigned char tileId = pOutData->getTile(x,y,z);
-					if (tileId > 0)
+                    if (tileId > 0 && tileId != 0xff)
 					{
-						if (currentLayer == 0 && Tile::tiles[tileId]->isEntityTile())
+                        if (currentLayer == 0 && Tile::tiles[tileId] && Tile::tiles[tileId]->isEntityTile())
 						{
 							shared_ptr<TileEntity> et = region.getTileEntity(x, y, z);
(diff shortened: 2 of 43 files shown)


codeHusky:
Tests fine, but resets FOV to 30 if you have existing settings prior to this change. Pls fix and then should be good to merge

jvnpr:
`eGameSetting_FOV` doesn't store the actual FOV, but rather a value from 0-100 which corresponds to the FOV slider's position. A value of 0 just corresponds to the slider minimum, 100 to the maximum, etc. Since the range of the FOV slider has been changed, the slider's saved position corresponds to a different FOV. You can test this by moving the slider to any given position in the newest nightly and moving the settings.dat file to a build of this branch.

To prevent a player's FOV from being changed when they update, I added the conversion check to see if we need to convert their old FOV setting to a new one. The way I implemented this check is by changing `eGameSetting_FOV` to have an offset, pushing the range to (101-201). This means that we can check if the player's `eGameSetting_FOV` is below the new minimum, and if it is, apply a conversion factor to set a new equivalent FOV in the new system.

PR #978, opened 2026-03-08
Chunk optimisations

Racc-Boi:
## **Description**

This PR introduces a multi-tiered optimization for the chunk rebuilding process. It replaces repeated branching logic with a pre-calculated lookup table and eliminates a redundant pre-calculation loop (removing a 3D boolean array). This significantly reduces memory bandwidth and CPU overhead on chunk rebuild threads, resulting in much faster terrain generation, quicker mesh building, and smoother rendering during player movement.

### **Previous Behavior**

During chunk rebuilding, the engine utilized a two-pass system:

1. **Pass 1 (Occlusion):** The engine iterated through a 16x16x16 grid, performing manual comparisons against a list of solid block types (e.g., stone, dirt, unbreakable, or ID 255\) for all neighbors to determine if a block was fully surrounded. It then stored the result in a bool isOccluded\[16\]\[16\]\[16\] array.  
2. **Pass 2 (Meshing):** The engine looped through the chunk *again*. If a block wasn't occluded, it had to manually check the neighbors a *second* time to figure out exactly which faces to render.

### **Root Cause**

* **Branching Overhead**: Modern CPUs are slowed down by "branching" (multiple if or || conditions) when doing manual block ID comparisons.  
* **Redundant Iterations & Memory**: The engine was unnecessarily initializing and looping through a 4,096-element (16x16x16) array every time a chunk was rebuilt.  
* **Double-checking Neighbors**: The engine evaluated the exact same block-solidness logic twice per visible block (once to check occlusion, once to cull faces).

### **New Behavior**

The engine now calculates occlusion purely inline using a single-pass system:

* **Lookup Table**: A pre-calculated boolean array isOccluder\[256\] is used to fetch a "True/False" solidness result using the Tile ID as a memory index (an O(1) operation).  
* **Inline Face Culling**: The isOccluded array and the first iteration pass have been completely removed.  
* **Reused Evaluations**: The engine calculates face visibility (drawLeft, drawRight, etc.) directly. If any face is visible, it renders the block and instantly reuses those boolean checks to build the specific faces, completely eliminating redundant neighbor checks.

### **Fix Implementation**

* **Tile Occlusion Lookup Table**: Implemented the isOccluder\[256\] lookup table centrally.  
* **Removed Redundant State**: Deleted bool isOccluded\[16\]\[16\]\[16\], saving 4KB of memory initialization per chunk rebuild and removing an entire nested for loop.  
* **Optimization Impact (Mathematical Breakdown)**:  
  * **Iteration**: \~2,744 interior tiles per chunk.  
  * **Old Method (Two Passes)**:  
    * Pass 1: 6 neighbor checks x 4 manual comparisons.  
    * Pass 2: 6 neighbor checks x 4 manual comparisons.  
    * Total: 2744 x 12 x 4 \= 131,712 branch comparisons per chunk.  
  * **New Method (Single Pass \+ Lookup Table)**:  
    * 6 face checks x 1 table lookup.  
    * Total: 2744 x 6 x 1 \= 16,464 table lookups per chunk.  
  * **Efficiency**: This represents an approximate 87.5% reduction in operations required for neighbor-occlusion and face-culling checks per chunk.  
* **Performance Gain**: By slashing CPU cycles and removing nested loops, chunk meshes are generated in a fraction of the time. This drastically reduces CPU spikes on the rebuild threads, meaning chunks appear on-screen visibly faster when generating new terrain or loading into the world.

Before: https://streamable.com/vl25rq
After: https://streamable.com/mazmot

### **AI Use Disclosure**

No AI was used

## **Related Issues**

N/A


Racc-Boi:
videos added

MarvelcoCode:
> videos added

can you try fixing stained glass not rendering water, its related to Chunk::rebuild() and LevelRenderer::render, ive already fixed it but broke ice blocks tiles

PR #980, opened 2026-03-08
texture expansion

ThatBlackAstronaut:
## Description
Expands the size of the texture maps to prepare for a possible parity to TU69 (update aquatic, 1.13)
## Changes
/Minecraft.Client/PreStitchedTextureMap.cpp : changed grid for how textures are mapped (1/16 of image to 1/32 of image)

expanded terrain.png (plus both mipmaps) and items.png to be larger for more texture space for future parity with where LCE was left off
### Previous Behavior
smaller texture map size, less room for textures

### Root Cause
not an issue other than being too small

### New Behavior
more space for textures

### Fix Implementation
/Minecraft.Client/PreStitchedTextureMap.cpp : changed grid for how textures are mapped (1/16 of image to 1/32 of image)

### AI Use Disclosure
none :D

## Related Issues

- Related to parity with this: https://minecraft.wiki/w/Xbox_360_Edition_TU69

quick edit: this is tested and everything works the same as before the change, other than it is able to hold more textures


Alexandra-Myers:
> Why merge at launch? Why not just fetch from each individual texture similarly to how Java does it?

Java doesn't do that... it stitches its texture atlases together on resource reload...

PR #981, opened 2026-03-08
Fix crash when launching at 720p or lower resolutions

### LINKED ISSUE #949: [Bug] Running the game at 1280x720 causes issues
### Bug description

Launching the game at 1280x720 (for example on a Steam Deck) causes a multitude of issues. I assume it's because the game assumes that its being ran on a PS3, maybe?... It causes issues like having the PS3 UI (as evident by the controller buttons) and some menus causing the game to crash, like the `Add server` menu.

### To Reproduce

1. Set your display resolution to 1280x720 (for example using gamescope on SteamOS)
2. Launch the game
3. Enjoy the bugs

### Expected behavior

* Not having the PS3 UI
* Not crashing when launching certain menus

### Screenshots

_No response_

### Videos

https://ptb.discord.com/channels/1478227187843858563/1478643695078211696/1480206092213682307 (from the Discord server)

### Version

dc91a1d

### Build Type

- [x] Release
- [ ] Debug

### Additional context

I'm running it on my Steam Deck through Proton Experimental, so I can't be entirely sure if it happens on Windows or not.

MrTheShy:
## Description

Fix crash on startup when the game window is at 720p or below on Windows64.

### Previous Behavior

Launching the game at 1280x720 or any resolution with height ≤720 caused an immediate crash during UI initialization. Iggy reported "Attempted to import undefined library platformskinHD.swf", followed by "Failed to load iggy scene!" and a `__debugbreak()` in `UIScene::loadMovie()`.

### Root Cause

In `UIController::loadSkins()`, the platform skin path and its Iggy registration name were chosen based on a runtime check (`m_fScreenHeight > 720.0f`). At 720p this evaluates to false, so the skin was loaded from `skinWin.swf` and registered as `"platformskin.swf"`. Below 720p, the same non-HD path was taken.

However, further down in the same function, the HD skin libraries (`skinHD.swf`, `skinHDHud.swf`, `skinHDInGame.swf`, etc.) are loaded **unconditionally** on Win64/Durango/Orbis — there is no resolution gate around them. These SWFs internally `import platformskinHD.swf`, which was never registered at ≤720p, causing Iggy to fail the import and every dependent scene to fail loading.

This is an original 4J bug — the HD block was written assuming these platforms always run above 720p, which is true for consoles but not for PC where the window can be any size.

### New Behavior

The game starts and runs correctly at any resolution on Windows64, including 720p and 480p.

### Fix Implementation

Removed the runtime resolution conditional for the platform skin path on Win64/Durango/Orbis. These platforms now always load the HD variant (`skinHDWin.swf` / `skinHDDurango.swf` / `skinHDOrbis.swf`) and always register it as `"platformskinHD.swf"`, matching what the unconditionally-loaded HD skin libraries expect. The registration was changed from a runtime `if` to a compile-time `#if` to make the platform split explicit. PS3/PSVita are unaffected as they have their own separate non-HD skin loading path.

### AI Use Disclosure

No AI was used to write the code in this PR.

## Related Issues
- Fixes #949

### DIFF
--- a/Minecraft.Client/Common/UI/UIController.cpp
+++ b/Minecraft.Client/Common/UI/UIController.cpp
@@ -529,43 +529,24 @@ void UIController::loadSkins()
 #elif defined __PSVITA__
 	platformSkinPath = L"skinVita.swf";
 #elif defined _WINDOWS64
-	if(m_fScreenHeight>720.0f)
-	{
-		platformSkinPath = L"skinHDWin.swf";
-	}
-	else
-	{
-		platformSkinPath = L"skinWin.swf";
-	}
+	// Windows64/Durango/Orbis always load HD skin libraries unconditionally,
+	// which import "platformskinHD.swf" by name.  The platform skin must
+	// therefore always be the HD variant regardless of screen resolution.
+	platformSkinPath = L"skinHDWin.swf";
 #elif defined _DURANGO
-	if(m_fScreenHeight>720.0f)
-	{
-		platformSkinPath = L"skinHDDurango.swf";
-	}
-	else
-	{
-		platformSkinPath = L"skinDurango.swf";
-	}
+	platformSkinPath = L"skinHDDurango.swf";
 #elif defined __ORBIS__
-	if(m_fScreenHeight>720.0f)
-	{
-		platformSkinPath = L"skinHDOrbis.swf";
-	}
-	else
-	{
-		platformSkinPath = L"skinOrbis.swf";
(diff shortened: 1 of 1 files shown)


codeHusky:
@MrTheShy We have multiple players reporting that at non-HD resolutions the HD textures are blurry. realistically you need to fix whatever the root cause was in sub-HD resolutions as iggy scales things super blurry with HD assets.
image
image

PR #998, opened 2026-03-09
Input: Add Ctrl+Q shortcut to drop entire item stack

### LINKED ISSUE #993: [Suggestion] CTRL Q to drop full stack
### Describe the solution you'd like

Just something that would be nice to have 

### Is your suggestion related to a problem?

_No response_

### Additional context

_No response_

Alezito2008:
## Description
Implements Ctrl+Q shortcut allowing players to drop entire item stacks at once

### Previous Behavior
Pressing the drop key only triggered the drop item action, dropping a single item regardless of any modifier keys.

### Root Cause
The input handling logic lacked modifier key detection for the drop action

### New Behavior
Pressing Q while holding Control now drops the entire stack of items

### Fix Implementation
- Input: Added `KEY_CONTROL` constant (mapped to VK_CONTROL) in KeyboardMouseInput.h.
- Logic: Updated the input tick to detect if Control is pressed when the drop action is active.
- Network: Refactored `MultiplayerLocalPlayer::drop(bool dropAll)` to switch between `DROP_ITEM` and `DROP_ALL_ITEMS` packet actions based on the boolean parameter.

### AI Use Disclosure
None

## Related Issues
- Fixes #993

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Minecraft.Client.exe -name Steve -fullscreen
 - **Inventory**: `E`
 - **Chat**: `T`
 - **Drop Item**: `Q`
+- **Drop Stack**: `Ctrl` + `Q`
 - **Crafting**: `C` Use `Q` and `E` to move through tabs (cycles Left/Right)
 - **Toggle View (FPS/TPS)**: `F5`
 - **Fullscreen**: `F11`


Persioqq:
Hello,

What do you think about changing this action from "Ctrl + Q" to "Hold Q" or "Hold B" (on the controller) to improve cross-platform support? We could also keep "Ctrl + Q" as an option for PC users. 

What are your thoughts?

Best regards.

PR #999, opened 2026-03-09
Revert the workaround of loading of hd textures

MrTheShy:
## Description

Fix blurry UI at sub-HD resolutions on Win64 by loading both HD and non-HD skin sets and picking the right scene SWF based on actual screen height.

### Previous Behavior

The earlier crash fix (from #981) forced HD skins at every resolution to prevent Iggy from failing to resolve `platformskinHD.swf`. This fixed the crash but at 720p and below Iggy was downscaling 1080p assets and everything looked blurry.

### Root Cause

Two issues, both from 4J's original code:

1. The HD skin libraries (`skinHD.swf`, `skinHDHud.swf`, etc.) are loaded unconditionally on Win64/Durango/Orbis. These SWFs internally `import platformskinHD.swf`. But the platform skin registration was behind a runtime resolution check — at <=720p it registered `skinWin.swf` as `platformskin.swf` instead of `platformskinHD.swf`, so Iggy couldn't resolve the import and crashed.

2. `loadMovie()` on Win64 was hardcoded to always load the 1080p scene SWF variant (`*1080.swf`), regardless of actual window size. The 1080p scenes import HD skins, so even if you fixed the platform skin you'd still be using HD assets at low resolutions.

4J actually started building a solution for this — they defined `eLibraryFallback_*` enum slots (including `eLibraryFallback_Platform`) and had code to load the non-HD skins into them, but it was gated behind `#ifndef _FINAL_BUILD` (debug only) and the platform skin slot was never actually loaded. So the fallback was incomplete and non-functional.

### New Behavior

The game starts and runs correctly at any resolution on Win64. Above 720p it uses 1080p scene SWFs with HD skins (same as before). At 720p and below it uses 720p scene SWFs with non-HD skins — native assets for the resolution, no downscaling, no blurriness. No crash at any resolution.

### Fix Implementation

**`UIController.h`**: Removed the `_FINAL_BUILD` preprocessor guard around the `eLibraryFallback_*` enum entries so they're available in release builds, not just debug.

**`UIController.cpp` — `loadSkins()`**: Rewrote the Win64 path to load both skin sets at startup:
- `eLibrary_Platform` = `skinHDWin.swf` registered as `platformskinHD.swf`
- `eLibraryFallback_Platform` = `skinWin.swf` registered as `platformskin.swf` (this was the 4J bug — slot existed but was never loaded)
- `eLibrary_*` = all HD skins (`skinHDGraphics.swf`, `skinHD.swf`, etc.)
- `eLibraryFallback_*` = all non-HD skins (`skinGraphics.swf`, `skin.swf`, etc.)

Both sets use distinct registered names so they coexist in the same Iggy session without conflicts.

**`UIScene.cpp` — `loadMovie()`**: On Win64 the scene SWF resolution is now chosen based on actual screen height instead of being hardcoded to 1080p:
- `>720p` loads `*1080.swf` (these import HD skins which import `platformskinHD.swf`)
- `<=720p` loads `*720.swf` (these import non-HD skins which import `platformskin.swf`)

The existing fallback (try 720 if 1080 file is missing) still works as a safety net.

### AI Use Disclosure
Ai was used to write this pr description.

## Related Issues
- Fixes the blurriness side effect from #981

### DIFF
--- a/Minecraft.Client/Common/UI/UIController.cpp
+++ b/Minecraft.Client/Common/UI/UIController.cpp
@@ -522,33 +522,9 @@ void UIController::tick()
 
 void UIController::loadSkins()
 {
-	wstring platformSkinPath = L"";
-
 #ifdef __PS3__
-	platformSkinPath = L"skinPS3.swf";
-#elif defined __PSVITA__
-	platformSkinPath = L"skinVita.swf";
-#elif defined _WINDOWS64
-	// Windows64/Durango/Orbis always load HD skin libraries unconditionally,
-	// which import "platformskinHD.swf" by name.  The platform skin must
-	// therefore always be the HD variant regardless of screen resolution.
-	platformSkinPath = L"skinHDWin.swf";
-#elif defined _DURANGO
-	platformSkinPath = L"skinHDDurango.swf";
-#elif defined __ORBIS__
-	platformSkinPath = L"skinHDOrbis.swf";
+	m_iggyLibraries[eLibrary_Platform] = loadSkin(L"skinPS3.swf", L"platformskin.swf");
 
-#endif
-	// Every platform has one of these, so nothing shared.
-	// On HD platforms (Win64/Durango/Orbis) the skin libraries always import
-	// "platformskinHD.swf", so we must register under that name at any resolution.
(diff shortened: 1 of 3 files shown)


codeHusky:
Tested and seems to work

PR #1002, opened 2026-03-09
Fixed game crash on connection error

tobleroneaddict:
## Description
Instead of the game entirely crashing on a connection issue it displays an in-game error instead

## Changes
Proper handling of invalid packets

### Previous Behavior
Game exception & crash

### Root Cause
Hacky Debug Assertions

### New Behavior
image

### Fix Implementation
Packet.cpp now handles null packets

### AI Use Disclosure
I used my head


tobleroneaddict:
D

Issue #267, opened 2026-03-03
Multiplayer Issues


glisteningtwilights:
Tried loading another instance just for fun and realized it wasn't working correctly, another player tries to connect and they just get stuck on the connecting screen.

However, they appear in the player list without being loaded in. Once they get thrown back to the menu they disappear from the list. Additionally their character appears on the world for a brief second before vanishing.

Another bug I noticed is if I close my world while someone is connecting my game crashes.

Side notes:
1. I noticed I had left online mode off and thought this was the issue but even after making sure online mode was on it still failed to connect.
2. I have only tested this with running another instance on my PC, I haven't tested from another device so this could just be an issue with running multiple instances but I'm not entirely sure.

![Image](https://github.com/user-attachments/assets/bda84200-fcfa-4512-b93c-b6cf25977a0c)
![Image](https://github.com/user-attachments/assets/45827c62-82ff-451c-9a85-eab381114013)

TimofeyLednev:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/175ca1e0-cfd4-4e98-b5d4-05205553bd36" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a9449d7d-2733-42b7-8bf8-09d89ec7aefc" />

![Image](https://github.com/user-attachments/assets/73ede01e-f537-4c34-b8cb-99b90cd7808a)

I have exactly the same problem: 1-2 screenshots of what I see, 3 screenshots of what my friend sees (my nickname is _MickPlay_YT_qiq and my friend's nickname is quant1k)

PR #1005, opened 2026-03-09
Fixed crash on joining a world

### LINKED ISSUE #345: Multiplayer Non-functional [CRASHES]
Tried to get my friend into a world with me and the game just crashes every time they try to join.

Image
Image

### LINKED ISSUE #784: [Bug] Failing to connect to a server causes game to lock up
### Bug description

If you're having any issue connecting to a server, or if it's taking a while, the game visibly freezes while it attempts to connect. This is due to the direct IP connection method locking the render thread rather than connecting in parallel. This looks like a crash due to it freezing after clicking "Join Game"

## Solution

We need to rework this code (PlatformNetworkManagerStub and related) to be called in a way that's more user friendly

tobleroneaddict:
## Description
Fixes the game crash on an invalid packet being received upon joining a world.

### Previous Behavior
Sometimes when joining the game would crash

### Root Cause
Debug assertions instead of proper handling

### New Behavior
No more crash

### Fix Implementation
Invalid packets are now routed back to the handler that exists in getTick()

### AI Use Disclosure
I used my head

## Related Issues
- Fixes #784 
- Fixes #345
- Related to #267

### DIFF
--- a/Minecraft.World/Packet.cpp
+++ b/Minecraft.World/Packet.cpp
@@ -327,28 +327,51 @@ shared_ptr<Packet> Packet::readPacket(DataInputStream *dis, bool isServer) // th
 	id = dis->read();
 	if (id == -1) return nullptr;
 
+	// Track last few good packets for diagnosing TCP desync
+	static thread_local int s_lastIds[8] = {};
+	static thread_local int s_lastIdPos = 0;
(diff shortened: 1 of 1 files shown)


tobleroneaddict:
Nevermind this is really annoying to get working on this version

Issue #1012, opened 2026-03-09
[Bug] Player joining server with the same UID causes bugs


codeHusky:
### Bug description

If two or more players join a server (LAN or otherwise) with the same UID, since a recent commit this is allowed behavior. This causes the players with the shared ID to be invisible and numerous unexpected behaviors.

MrTheShy:
it's because of a kick the old player measure i took, but yeah, can be totally reverted there's virtually no reason for it to be there

PR #1013, opened 2026-03-09
Reject duplicate UIDs on login and remove noisy gdraw debug log

MrTheShy:
## Description

Prevents players from joining a server if their UID is already in use by another connected player, fixing invisible players and undefined behaviour caused by duplicate UIDs coexisting in the player list.

### Previous Behavior
When a player tried to join a server with a UID matching an already-connected player, the server would force-disconnect the existing player and accept the new connection. This allowed two clients with the same UID to briefly or permanently coexist in the player list, resulting in invisible players and other undefined behaviour.

### Root Cause
PR #767 introduced a reconnect-handling code path in `PendingConnection::handleLogin` that, upon detecting a duplicate UID, would force-kick the stale player and call `handleAcceptedLogin` for the incoming connection. While intended for reconnect scenarios, this unconditionally allowed any duplicate-UID connection through at the cost of evicting the legitimate existing player.

### New Behavior
When a duplicate UID is detected during login, the incoming connection is disconnected. The existing player remains on the server unaffected.

### Fix Implementation
In `PendingConnection::handleLogin`, the `else if (duplicateXuid)` branch previously force-disconnected the stale player, recycled their `smallId`, and called `handleAcceptedLogin` for the new connection. This block has been replaced with a simple `disconnect(DisconnectPacket::eDisconnect_Banned)` call on the incoming connection, consistent with how duplicate names are already handled under `g_bRejectDuplicateNames`.

A leftover per-frame `DebugPrintf` in `gdraw_SetViewSizeAndWorldScale` (introduced during the resolution-fix work in #767) is also removed.

## AI Use Disclosure
No AI was used to write the code in this PR.
AI fully wrote this PR description

## Related Issues
- Related to #767 
- Related to #1012

### DIFF
--- a/Minecraft.Client/PendingConnection.cpp
+++ b/Minecraft.Client/PendingConnection.cpp
@@ -190,34 +190,11 @@ void PendingConnection::handleLogin(shared_ptr<LoginPacket> packet)
 	}
 	else if (duplicateXuid)
 	{
-		// The old player is still in PlayerList (disconnect hasn't been
-		// processed yet).  Force-close the stale connection so the
-		// reconnecting client isn't rejected.
-		app.DebugPrintf("RECONNECT: Duplicate xuid for name: %ls, forcing old connection closed\n", name.c_str());
-		shared_ptr<ServerPlayer> stalePlayer = server->getPlayers()->getPlayer(loginXuid);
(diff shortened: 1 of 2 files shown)


MrTheShy:
@codeHusky hope this doesn't fall in the not do two things at once, just removing some logging i missed earlier, spams the console.

PR #1017, opened 2026-03-09
Fix font rendering for color and formatting codes 

ayushthoren:
**Edits: This PR was originally made to fix just the "Colormatic" splash text appearance, but has since fixed more general issues with the font renderer.**

## Description
Fixes issues with font color and formatting when using the tesselator-based text renderer.

### Previous Behavior
Text using formatting codes such as color (`§c`, `§a`) and styles (`§n`, `§m`, etc.) could render incorrectly.
- Colored segments would end up overwriting previous color formatting for parts of the string in the same batch.
- Underline and strikethrough lines sometimes appeared with the wrong color or did not match the text color.

Notably, one consequence of this is that the Colormaitc splash (`§1C§2o§3l§4o§5r§6m§7a§8t§9i§ac`) rendered in one flat color instead of showing per-segment colors.

### Root Cause
Font rendering mixed two rendering paths:
- Glyphs were emitted through the tesselator batch.
- Underline and strikethrough lines were drawn using a helper that relied on GL color state.
Because the style lines were not part of the tesselator batch, they did not use the same vertex color as the text and could render incorrectly.

### New Behavior
Formatted text now renders consistently, with underline and strikethrough lines matching the color of the surrounding text and all formatted glyphs rendering correctly in the same batch.

### Fix Implementation
In `Minecraft.Client/Font.cpp`, style line rendering was moved into the tesselator path.
- Underline and strikethrough quads are now emitted directly during glyph rendering.
- A small helper (`addSolidQuad`) was added to emit these quads inside the active tesselator batch.
- Glyph vertex emission was moved into `emitCharacterGeometry` to remove duplicated geometry code.

This ensures previous glyphs keep their original color and new ones use the updated color.

### AI Use Disclosure
I used AI as a tool to track down the possible root causes of the rendering issues, then confirmed them and implemented the fixes myself.

## Related Issues
- Fixes Discord Thread 1480350614977384550

## How it looked before:
**Test splash**
Screenshot 2026-03-09 192732
**Colormatic splash**
image

## How it looks after:
**Test splash**
Screenshot 2026-03-09 195611
**Colormatic splash**
image

### DIFF
--- a/Minecraft.Client/Font.cpp
+++ b/Minecraft.Client/Font.cpp
@@ -45,46 +45,46 @@ Font::Font(Options *options, const wstring& name, Textures* textures, bool enfor
 	random = new Random();
 
 	// Load the image
-    BufferedImage *img = textures->readImage(textureLocation->getTexture(), name);
+	BufferedImage *img = textures->readImage(textureLocation->getTexture(), name);
 
 	/* - 4J - TODO
 	try {
-        img = ImageIO.read(Textures.class.getResourceAsStream(name));
-    } catch (IOException e) {
-        throw new RuntimeException(e);
-    }
+		img = ImageIO.read(Textures.class.getResourceAsStream(name));
+	} catch (IOException e) {
+		throw new RuntimeException(e);
+	}
 	*/
 
-    int w = img->getWidth();
-    int h = img->getHeight();
-    intArray rawPixels(w * h);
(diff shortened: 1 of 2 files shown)


ayushthoren:
> any updates? I'd love to get this fix in :)

Oops, I should have been more clear with my previous comment. My last commit has fixed those issues (according to my testing). So if it appears to work on your end, it should be good to merge! :)

PR #1025, opened 2026-03-09
Netherportal does not light if an extra obsidian is next to the fire

### LINKED ISSUE #960: [Bug] Netherportal does not light if an obsidian is next to the frame
### Bug description

Netherportal does not light if there is an obsidian (not from the frame itself) next to the fire (from flint and steel for example) (see picture)

### To Reproduce

1. Build obsidian frame
2. Place 1 obsidian next to where you are going to use flint and steel

### Expected behavior

Nether portal lights and you can enter it.

### Screenshots

Image

### Videos

_No response_

### Version

dc91a1d

### Build Type

- [x] Release
- [x] Debug

### Additional context

_No response_

Us3ful-dev:
## Description
netherportal does not light if the fire (from flint and steel for example) is next to an extra block of obsidian (see picture).
559935838-e8db5fd9-3cd7-4d2b-9ed1-8f7062cb1245

!!!IMPORTANT: The old behaviour was apparently a feature (see issue #960) but i find it very annoying and think it should be fixed (found it while speedrunning).

## Changes
Added function to check for valid portalframes and posibly spawn the portal blocks if needed.
Added check for two posible portals.

### Previous Behavior
The portal does not light and the fire remains.

### Root Cause
If there is an obsidian in the x and z direction it always returns false.

### New Behavior
Light one of the portals (prioritizes the x direction) or just the frame if it is an extra obsidian next to the fire.

### Fix Implementation
Check if there is an obsidian in the x and z direction, if so check both sides for a portal.

### AI Use Disclosure
No AI was used

## Related Issues
- Fixes #960


CL0utM4N:
Closed VS 2022, merged your pull request into the latest build and and rebuilt but didn't seem to work 

Screenshot 2026-03-11 153017

https://github.com/user-attachments/assets/b251283c-99c9-40d8-a58a-7f135f16ac38

Issue #1022, opened 2026-03-09
[Bug] Splitscreen hints either show up squished or dont appear at all


IsaacWarhol:
### Bug description

When playing split screen hint pop ups appear squished when in vertical splitscreen mode and dont appear at all when in horizontal split screen mode.

### To Reproduce

Enter a world and begin splitscreen
Open crafting menu and attempt to craft something without the required materials.

### Expected behavior

The pop up hint should appear and not be squished in any way

### Screenshots

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/32766fb4-84cc-4db6-b9ed-dff10e4e8d66" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/971ab9d5-f8f5-455e-bf27-7faf374843a7" />

### Version

7a4f57e

### Build Type

- [x] Release
- [ ] Debug

PR #1031, opened 2026-03-09
Fix multiplayer non host split screen

MrTheShy:
## Description

Several split-screen bugs affecting the non-host player on Windows64: joining a session started via the UI, phantom joins from unfocused windows, tutorial popup misplacement, and the debug overlay being clipped to one viewport. This PR addresses all four.

### Previous Behavior

- **Split-screen join via UI**: When the host connected to a server through the in-game UI (rather than the command line), the `g_Win64MultiplayerIP` / `g_Win64MultiplayerPort` globals were never populated. A second local player attempting to join split-screen would therefore try to connect to an empty address, failing silently.
- **Phantom joins on unfocused window**: Gamepad button state was polled every tick regardless of whether the game window had focus. Any gamepad input processed while the user was in a different application could inadvertently trigger a split-screen player join.
- **Tutorial popup positioning**: The popup's render position was computed via a hard-coded `switch` over viewport types, manually offsetting by `screenWidth/2` or `screenHeight/2`. This did not account for the actual 16:9 fit box used by every other scene, so the popup appeared at the wrong coordinates — especially visible in quadrant and non-square split layouts.
- **Debug overlay in split-screen**: Opening the debug overlay (F3) during split-screen opened it in the active player's viewport group, clipping it to half the screen.

### Root Cause

- **Join via UI**: `WinsockNetLayer::JoinGame` (the path taken by a UI-initiated connection) did not write the resolved IP and port back into the globals that `JoinSplitScreen` reads. The globals remained zeroed-out from startup.
- **Phantom joins**: The `tryJoin` condition in `Minecraft::run_middle` had no window-focus guard, so gamepad input was evaluated even when the application was not in the foreground.
- **Tutorial popup**: The manual coordinate table was duplicating — incorrectly — the viewport-to-screen-space mapping that `UISplitScreenHelpers` already provides. It also passed the raw viewport dimensions to `IggyPlayerSetDisplaySize`, ignoring the fitted sub-rectangle.
- **Debug overlay**: `ui.NavigateToScene` was called without an explicit UI group, defaulting to the current split-screen group instead of the fullscreen group.

### New Behavior

- A second local player can join split-screen successfully even when the host connected through the UI.
- Gamepad input is ignored for the join trigger while the game window is not focused, preventing phantom joins.
- The tutorial popup renders at the correct position and size inside each split-screen viewport, consistent with all other UI scenes.
- The debug overlay spans the entire window when opened during split-screen, making all debug information readable regardless of the current layout.

### Fix Implementation

**`WinsockNetLayer::JoinGame`** (`WinsockNetLayer.cpp`): After a successful connection and small-ID assignment, copy the resolved `ip` and `port` into `g_Win64MultiplayerIP` / `g_Win64MultiplayerPort`. Added a guard in `PushFreeSmallId` to never recycle small IDs in the range `[0, XUSER_MAX_COUNT)`, which are permanently reserved for the host's local pads — this prevents a recycled ID collision when a remote client reconnects.

**`Minecraft::run_middle`** (`Minecraft.cpp`): Appended `&& g_KBMInput.IsWindowFocused()` to the `tryJoin` boolean so the join path is only evaluated when the window has focus.

**`UIComponent_TutorialPopup::render`** (`UIComponent_TutorialPopup.cpp`): Replaced the manual per-viewport coordinate switch with `GetViewportRect` + `Fit16x9` from `UISplitScreenHelpers.h` (the same helpers used by `UIScene::render`). The fitted width/height are now also passed to `IggyPlayerSetDisplaySize` instead of the raw viewport dimensions. Added `default: break` to both safezone switches to silence compiler warnings.

**`Minecraft::tick`** (`Minecraft.cpp`): Pass `eUIGroup_Fullscreen` as the fifth argument to `ui.NavigateToScene` when opening the debug overlay, so it is placed in the fullscreen group and covers the whole window.

### AI Use Disclosure

Was used to write this pr description 

## Related Issues
- Related to #1022  and others

### DIFF
--- a/Minecraft.Client/Gui.cpp
+++ b/Minecraft.Client/Gui.cpp
@@ -443,7 +443,8 @@ void Gui::render(float a, bool mouseFree, int xMouse, int yMouse)
 				double maxHealth = minecraft->localplayers[iPad]->getAttribute(SharedMonsterAttributes.MAX_HEALTH);
 
 				double totalAbsorption = minecraft->localplayers[iPad]->getAbsorptionAmount();
-				int numHealthRows = Mth.ceil((maxHealth + totalAbsorption) / 2 / (float) NUM_HEARTS_PER_ROW);
+				const double healthHalves = (maxHealth + totalAbsorption) / 2.0;
+				int numHealthRows = Mth.ceil(healthHalves / (float) NUM_HEARTS_PER_ROW);
 				int healthRowHeight = Math.max(10 - (numHealthRows - 2), 3);
 				int yLine2 = yLine1 - (numHealthRows - 1) * healthRowHeight - 10;
 				absorption = totalAbsorption;
@@ -469,7 +470,7 @@ void Gui::render(float a, bool mouseFree, int xMouse, int yMouse)
 				}
 
 				//minecraft.profiler.popPush("health");
-				for (int i = Mth.ceil((maxHealth + totalAbsorption) / 2) - 1; i >= 0; i--)
+				for (int i = (int)Mth.ceil(healthHalves) - 1; i >= 0; i--)
 				{
 					int healthTexBaseX = 16;
 					if (minecraft.player.hasEffect(MobEffect.poison))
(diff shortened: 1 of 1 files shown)


MrTheShy:
@codeHusky ready to be reviewed, fixed every issue i could find, let me know if this can be merged soon because i have to build other stuff based on this and wouldn't be clean to make another branch based on this before it gets merged. sylvessa [tested it](https://discord.com/channels/1478227187843858563/1479662210295922750/1480691904294293694) and i've also deeply tested this

PR #1065, opened 2026-03-09
Add cancel join functionality

irice7:
## Description
Very minor QoL addition to allow cancelling while joining a server (came from the feature-suggestions Discord channel thread 1480269665908555919)

### Previous Behavior
Previously when joining a server you have no way of stopping and getting out of the fullscreen progressbar.

### Root Cause
If connecting to a server takes a long time you need to wait until the connection fails.

### New Behavior
Adds the Cancel Join tooltip and behavior, allowing you to close the connection deliberately.

### Fix Implementation
Uses a boolean (`m_bCancelRequested`) check in the loop that establishes the connection with the server within `GameNetworkManager.cpp`, where the main thread can set the boolean to true. Causing the connection to close.
Will also hide the tooltip when it is pressed.

### AI Use Disclosure
AI was used as a rubber duck to fix a crash

## Related Issues
None

image


irice7:
> Does this also resolve the failed connections freezing the whole game?

Doesn't happen to me. How do you trigger this bug?

PR #1089, opened 2026-03-10
Fix issue where visually the HUD shows you have 0 hearts when you dont

### LINKED ISSUE #753: [Bug] 0 hearts and still alive
### Bug description

Iam on 0 hearts and i didn't die
Image
I got hit by a venenous spider (the one that applies poison) and i think i got hit at the same time by a regular attack from one of them resulting in me didn't dead and stuck on 0 heart
Luckliy if you eat anything you can heal back 
Image

### To Reproduce

1. Wait for a spider to hit and applies poison
2. Get hit by a regular attack from them at the same time of the poison tick to apply

### Expected behavior

Umm I should've died there

### Screenshots

_No response_

### Videos

_No response_

### Version

latest

### Build Type

- [x] Release
- [ ] Debug

### Additional context

_No response_

3UR:
## Description
This is something 4J Studios fixed in TU20 (https://minecraft.fandom.com/wiki/Xbox_360_Edition_TU20)
image

4J also used ceil() in TU20 so this code is the exact same changes they made
image

### Previous Behavior
HUD would display you have 0 hearts when you have a half of a heart.

### Root Cause
The game was casting the float to a int which truncated the decimal.

### New Behavior
HUD will now show half of a heart instead of none.

### Fix Implementation
Use ceil() to prevent the decimal being truncated.

### AI Use Disclosure
No AI was used.

## Related Issues
- Fixes #753
- Related to #753

### DIFF
--- a/Minecraft.Client/Common/UI/IUIScene_HUD.cpp
+++ b/Minecraft.Client/Common/UI/IUIScene_HUD.cpp
@@ -195,8 +195,8 @@ void IUIScene_HUD::renderPlayerHealth()
 	// Update health
 	bool blink = pMinecraft->localplayers[iPad]->invulnerableTime / 3 % 2 == 1;
 	if (pMinecraft->localplayers[iPad]->invulnerableTime < 10) blink = false;
-	int currentHealth = (int)ceil(pMinecraft->localplayers[iPad]->getHealth());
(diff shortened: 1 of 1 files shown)


codeHusky:
should use the modern style casts in-line with the rest of the project (static_cast and whatnot)

PR #1090, opened 2026-03-10
Fix: Music selection + Critical hit implementation

tylerreese543-lab:
## Description

This PR builds upon the SoundEngine::getMusicID() method to be better.
It also includes proper implementation of the critical hit sound into the game.
## Changes

Added New Ranges within the SoundEngine to account for new Menu and Creative ranges.
Re-located the "piano" enums in the eStream_Overworld enumerator to be BEFORE creative and menu music, separating the two.
Added capability for these two ranges to be added and initialized in SoundEngine::SetStreamingSounds()
Queried the state of the game (world loaded, player 1 gamemode) to properly play menu, creative, or survival music with menu being the default.

Adds the critical.mp3 sound file into the game. 
Added the eSoundType/wchSoundNames enumerators which point to that file.
Added new sound functionality (getCritHurtSound) alongside "getHurtSound" to properly retrieve the sound enumerator.

Added an "isCritical" flag to Damage Sources. Added proper getter/setter methods for such.

Added a new EntityEvent enumerator for critical hits. Added functionality within the LivingEntity::handleEntityEvent() override to get proper crit sounds for players.
### Previous Behavior

Would just simply play the overworld range which contained creative/menu sounds

Critical PARTICLES showed, but not the actual sound.

Would simply not play any critical hit sound; would just play the normal "hurtfleshn" sounds.
### Root Cause

The Enumerator range did not account for these things in any respect. It simply included the creative and menu music as normal "overworld" tracks.

LivingEntity::hurt had zero differentiation between normal and critical hit sounds, but particles could. There was no way to tell if a damage source came from a critical attack and thus no way to properly play the sound.

There was no EntityEvent enumerator for a critical hit. Simple entities (normal mobs) did not use this, so they worked fine. Players did because they operated in different functions, so there was no code that played the critical hit sound nor was there a check for it.
### New Behavior

Now, the code has more proper discernment and awareness of the state of the game. It plays the correct music accordingly.

Now the sound plays depending on whether the attack was a critical one.

Now the code can differentiate between a hurt and critical hurt sound and play the right sounds properly for players.
### Fix Implementation

I added query points for player 1's gamemode and general existence; for some reason "pMinecraft == nullptr" was insufficient. It now defaults to the menu music; mash-up packs also are bound to the same ranges because they dont have proper formatting 😢 

In the Player::attack function, the "isCritical" flag in the DamageSource is set to true, if the function's "bCrit" value is (being the same value used to determine particles). 
This flag is then retrieved from LivingEntity::hurt, and will use the new "getCritHurtSound()" virtual function to retrieve eSoundType_DAMAGE_CRITICAL, which points to the new critical.mp3 sound file and plays it correctly

In the EntityEvent class, I added CRITICAL_HURT, which functions pretty much exactly like hurt, except it acts as a flag to play the critical hit sound when LivingEntity::handleEntityEvent is called.

Changed the code in the LivingEntity::hurt() function to check for source->isCritical, and sends a different event ID depending on the critical status.
### AI Use Disclosure

No AI was used here.
## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]


codeHusky:
Fill out the PR template or this will be closed

PR #1091, opened 2026-03-10
Fix 3 RCEs in packet parsing and texture loading

JuelzIrons:
## Description
This PR fixes 2 RCE vulnerabilities in Packet Parsing and 1 in texture loading

### Previous Behavior
packets with crafted RoomSyncData or AddPlayerFailed sizes had no validation, allowing integer overflow and unbounded heap allocation and wsprintfW in AbstractTexturePack.cpp had no size limit on the output buffer.

### Root Cause
Old code had no bounds checking on network packet sizes, and wsprintfW had no buffer size parameter.

### New Behavior
Packet sizes are validated and capped before allocation. swprintf_s enforces buffer size on string formatting.

### Fix Implementation
`wsprintfW` replaced with `swprintf_s` with `LOCATOR_SIZE` in AbstractTexturePack.cpp
RoomSyncData and AddPlayerFailed packet parsing now reads size into `receivedSize`, rejects if `< 0 || > 100000`, checks `totalSize` for integer overflow before allocating

### AI Use Disclosure
No AI was used in the discovery or patching of these RCEs

## Related Issues
N/A

### DIFF
--- a/Minecraft.Client/Durango/Network/DQRNetworkManager_SendReceive.cpp
+++ b/Minecraft.Client/Durango/Network/DQRNetworkManager_SendReceive.cpp
@@ -217,7 +217,8 @@ void DQRNetworkManager::BytesReceivedInternal(DQRConnectionInfo *connectionInfo,
 						memcpy( &receivedSize, connectionInfo->m_pucRoomSyncData, 4);
 						delete [] connectionInfo->m_pucRoomSyncData;
 
-						if( receivedSize < 0 || receivedSize > 1048576 )
+					// a 512KB limit gives enough headroom for legitimate XUID data, but prevents a overflow. With 2 players the size is about 100kb so 512kb gives good headroom
+						if( receivedSize < 0 || receivedSize > 524288 )
 						{
 							DQRNetworkManager::LogCommentFormat(L"RoomSyncData rejected: size=%d", receivedSize);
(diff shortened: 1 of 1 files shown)


Minecraft.Client/Durango/Network/DQRNetworkManager_SendReceive.cpp
```diff
@@ -296,8 +317,21 @@ void DQRNetworkManager::BytesReceivedInternal(DQRConnectionInfo *connectionInfo,
 					// At first stage of reading the 4 byte count
 					if( connectionInfo->m_addFailedPlayerDataBytesRead == 4 )
 					{
-						memcpy( &connectionInfo->m_addFailedPlayerDataBytesToRead, connectionInfo->m_pucAddFailedPlayerData, 4);
+						int receivedSize = 0;
+						memcpy( &receivedSize, connectionInfo->m_pucAddFailedPlayerData, 4);
 						delete [] connectionInfo->m_pucAddFailedPlayerData;
+
+						if( receivedSize < 0 || receivedSize > 1048576 )
```
codeHusky: same here

JuelzIrons:
I decreased the limit to 512kb + added comments as to why its 512kb
1mb was more or so a miscalculation on my end, 

512 is still on the forgiving side still but it still prevents the overflow

PR #1104, opened 2026-03-10
Add smoothBlend functionality to crispBlend for transparent texture mipmaps

TorutheRedFox:
## Description
Fixes water opacity in default texture pack, making it no longer fully opaque when mipmaps kick in.

### Previous Behavior
Water was fading to opaque due to crispBlend destroying texture alpha in mipmaps.

### Root Cause
This seems to be a long standing bug in LCE because they used Notch's crispBlend algorithm to blend texels when calculating mipmaps in realtime, which is intended for cutout textures, not transparent textures. Java edition scrapped the generation of mipmaps for a good while, but in its unused implementation it used smoothBlend to calculate the mips for textures that were supposed to be transparent, i.e. water. This PR just merges their functionality based on the detected alpha value of the pixels.

### New Behavior
Transparent textures such as water keep their transparency in their mipmaps.

### Fix Implementation
If both pixels involved in the blend operation have an alpha value less than 224, they'll be treated as transparent and the smoothBlend algorithm is used to blend the pixels instead, preserving the alpha in the process. If either pixel has an alpha value of 224 or more, crispBlend blends them as normal to avoid losing detail in cutout textures.

### AI Use Disclosure
No AI was used in the making of this code.

### DIFF
--- a/Minecraft.Client/Texture.cpp
+++ b/Minecraft.Client/Texture.cpp
@@ -716,38 +716,47 @@ int Texture::crispBlend(int c0, int c1)
 	int a0 = static_cast<int>(((c0 & 0xff000000) >> 24)) & 0xff;
 	int a1 = static_cast<int>(((c1 & 0xff000000) >> 24)) & 0xff;
 
-	int a = 255;
-	if (a0 + a1 < 255)
+	// continue with crisp blend if it's likely to be an opaque/cutout tile in the atlas
+	if (a0 >= 0xfa || a1 >= 0xfa)
 	{
-		a = 0;
-		a0 = 1;
-		a1 = 1;
+		int a = 255;
+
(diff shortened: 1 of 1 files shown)


mattsumi:
> when i get home :P

I'd love to see this if this is still available.

GabsPuNs:
I got it working in my fork, like a graphics slider, Potato, Fast, Fancy and Extra.

Potato
image
Fast
image
Fancy
image
Extra
image
Also this is how the water looks when is loading in extra lol
image

TorutheRedFox:
the water has mipmaps either way, they're just generated in realtime using a function that intentionally destroys alpha information as it was designed for the opaque cells on the atlas

Notch intended smoothBlend to be used for transparent stuff like ice and water

PR #1135, opened 2026-03-11
Fix FOV Slider + Viewmodel

### LINKED ISSUE #655: [Bug] Hand FOV changes when the FOV is changed in graphics settings

jvnpr:
# [!] This PR is dependent on #1097 

## Description
Change the FOV slider to use the existing fov option system instead of modifying the base FOV value in `gameRenderer`. 
Expands the slider range from 70-110 to 30-110, and makes slider values 1:1 with ticks. 

### Previous Behavior
The FOV slider values are not accurate compared to other platforms and causes issues when high FOV values are used. FOV setting also affects the hand viewmodel.

lcenightlybuild110 java110

### Root Cause
Current implementation uses the `gameRenderer` function `SetFovVal()`, which changes the base FOV value used to render the game. `gameRenderer` uses this base FOV for both the screen and hand viewmodel.

The slider was implemented by creating a slider with a range of 0-100, then mapping that range to 70-110. This means the slider must be moved multiple ticks to change the value by 1. 

### New Behavior
Switched to using the method `Minecraft::GetInstance()->options->fov` to adjust FOV.
Hand viewmodel no longer is affected by FOV setting.
Slider values are now 1:1 with ticks. 

fovsliderfix110java110

fovsliderfix30java30

### Fix Implementation
Change `CMinecraftApp::ActionGameSettings()` and slider implementations to use the method `Minecraft::GetInstance()->options->fov` to adjust FOV.

Change slider range to 0-80 instead of 0-100, change `eGameSetting_FOV` to store FOV as a value 0-80 corresponding to the range 30-110 instead of a value 0-100

Use `SettingFixer()` (from #1097) to seamlessly convert old saved settings to new format.

### AI Use Disclosure
No AI was used to author these changes or pull request.

## Related Issues
- Fixes #655, #666, #771, #838, et al. 
- Related to #1097, #975

### DIFF
--- a/Minecraft.Client/Common/App_enums.h
+++ b/Minecraft.Client/Common/App_enums.h
@@ -178,6 +178,9 @@ enum eGameSetting
 	// PSVita
 	eGameSetting_PSVita_NetworkModeAdhoc,
 
+	// Add setting data version
+	eGameSetting_SettingDataVersion,
+
 
 };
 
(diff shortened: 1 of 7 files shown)


codeHusky:
Already fixed by another PR

PR #1138, opened 2026-03-11
Fix: Creative/Survival/Menu Music Issues

### LINKED ISSUE #1301: [BugCreative music plays in survival
### Bug description

For some reason sometimes music which should play only in creative mode starts playing in survival, kinda wierd

### To Reproduce

Not sure how to reproduce, happens spontaniosly

### Expected behavior

Creative music playing in survival

### Screenshots

_No response_

### Videos

_No response_

### Version

latest

### Build Type

- [ ] Release
- [ ] Debug

### Additional context

_No response_

tylerreese543-lab:
## Description

Separation from previous closed PR #1090 

Separates music properly and prevents music from the wrong gamemode/state playing.
### Previous Behavior

Sometimes the incorrect songs would play as there was no differentiation of the game's states (eg survival songs might play in the menu, and menu songs might play in creative mode).
### Root Cause

This was because 1. The "piano3" enumerator that the SoundEngine uses came AFTER all of the creative/menu songs, so they were included in the range of songs available to play.
### New Behavior

Now the code queries the state of the game and the gamemode of the primary player to determine music properly.
### Fix Implementation

Added two new [int] ranges for creative and menu music
*, **Updated the SoundEngine::SetStreamingSounds to have arguments for these two ranges
Queried the player's gamemode to differentiate between creative and survival music
All other music defaults to the menu music if no player is present.

*Mash-up packs don't differentiate between these two ranges, so they are identical in the DLCTexturePack.cpp file.
**Defaulted to Creative1-6, Menu1-4.
### AI Use Disclosure

No AI was used 🫡 
## Related Issues
- Fixes #1301 
- Related to #1090

### DIFF
--- a/Minecraft.Client/Common/Audio/SoundEngine.cpp
+++ b/Minecraft.Client/Common/Audio/SoundEngine.cpp
@@ -114,7 +114,9 @@ const char *SoundEngine::m_szStreamFileA[eStream_Max]=
 	"hal4",
 	"nuance1",
 	"nuance2",
-
+	"piano1",
+	"piano2",
+	"piano3",
 #ifndef _XBOX
 	"creative1",
 	"creative2",
@@ -127,11 +129,6 @@ const char *SoundEngine::m_szStreamFileA[eStream_Max]=
 	"menu3",
 	"menu4",
 #endif
-
-	"piano1",
-	"piano2",
-	"piano3",
-
 	// Nether
 	"nether1",
 	"nether2",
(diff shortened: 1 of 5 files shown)


tylerreese543-lab:
I'm all for re-factoring but this seems to add too much functionality to something that doesn't need it

Even when mash-up packs are added you just plug the new ranges into SetStreamingSounds

+ when mini-games are added you can do the same thing

PR #1141, opened 2026-03-11
Fix: Critical Hit Sound Implementation

tylerreese543-lab:
## Description

Implements the critical hit sound for both players and mobs.
## Changes

Added the "eSoundType_DAMAGE_CRITICAL" enum.
Added LivingEntity::getCriticalSound() which returns an eSoundType, being the DAMAGE_CRITICAL variant.
Added the EntityEvent::HURT_CRITICAL event enumerator.
Added critical.mp3 to the damage sounds.
Added functionality for critical hits to be played.
Added the "_isCritical" flag for DamageSources.
Changed LivingEntity::handleEntityEvent to account for critical hits.
### Previous Behavior

No critical hit was programmed prior, would display particles but no sound would play.
### Root Cause

There was no way to differentiate between critical and normal hits when taking actual damage and playing sound.
### New Behavior

Now, DamageSources can carry whether the attack was a critical one or not. Plays a sound accordingly
Sends EntityEvent::HURT or EntityEvent::HURT_CRITICAL depending on the DamageSource data
Uses the getCriticalSound() function to retrieve the proper sound and plays it alongside the getHurtSound().
### Fix Implementation

DamageSource::setisCritical() is called based on the "bCrit" variable in Player::attack(), before sending it to "LivingEntity::hurt()"
Uses DamageSource::isCritical() to check whether the attack was a critical one.
This is used to determine which EntityEvent to broadcast.
This is also used to determine whether to play the critical sound in "LivingEntity::hurt()"
### AI Use Disclosure
No AI

## Related Issues
- Fixes #[issue-number]
- Related to #1090


Minecraft.Client/Common/BuildVer.h
tylerreese543-lab: damnit.

codeHusky:
Merged outside of normal order outlined by Contributing.md primarily for "correctness" - the lack of a sound for crits is very inconsistent with every other edition of the game and _feels_ buggy. Fixed in [TU46](https://minecraft.wiki/w/Xbox_360_Edition_TU46#General_2) officially

tylerreese543-lab:
Created a new PR that fixes the original .mp3 to a .ogg like the other sounds

PR #1149, opened 2026-03-11
Fix: Pose Height/Animation Fixes

tylerreese543-lab:
## Description

Fixes the sneak height so the player doesn't float when they sneak. I might merge this with other fixes like boats in the future.

Fixes the boat height / animation for both large and small models.
### Previous Behavior

Player floated when they crouched.

Player floated on boat, and legs were together.

### Root Cause

Their height in the render script did not go down upon sneak.

Simply did not have the animation code proper.
### New Behavior

Added that functionality.

Fixed the improper animation code. Sitting/Riding animations work properly, and small models are positioned correctly.

image
image
image

### Fix Implementation

Replaced "yp" in PlayerRenderer::render, to go down upon sneak. I'm not entirely sure why it blocked sneaking in the first place.

Added a new bitmaskEnabled() function for getBitmaskOverride [localized, cleaned up, readability]
Re-worked boat riding
Added translation for item holding when small models are in a boat.
## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]

### DIFF
--- a/Minecraft.Client/HumanoidModel.cpp
+++ b/Minecraft.Client/HumanoidModel.cpp
@@ -140,6 +140,10 @@ HumanoidModel::HumanoidModel(float g, float yOffset, int texWidth, int texHeight
 	_init(g,yOffset,texWidth,texHeight);
 }
 
+bool HumanoidModel::bitmaskEnabled(unsigned int uiBitmask, animbits bit)
+{
+	return (uiBitmask&(1<<bit))>0;
+}
 void HumanoidModel::render(shared_ptr<Entity> entity, float time, float r, float bob, float yRot, float xRot, float scale, bool usecompiled)
 {	
 	if(entity != nullptr)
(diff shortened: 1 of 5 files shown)


Minecraft.Client/HumanoidModel.cpp
```diff
@@ -193,30 +198,30 @@ void HumanoidModel::setupAnim(float time, float r, float bob, float yRot, float
...
-		else if(uiBitmaskOverrideAnim&(1<<eAnim_ArmsOutFront))
+		else if(bitmaskEnabled(uiBitmaskOverrideAnim, eAnim_ArmsOutFront))
 		{
 			arm0->xRot=-HALF_PI;
 			arm1->xRot=-HALF_PI;
 			arm0->zRot = 0.0f;
 			arm1->zRot = 0.0f;
 		}
-		else if(uiBitmaskOverrideAnim&(1<<eAnim_SingleArms))
+		else if(bitmaskEnabled(uiBitmaskOverrideAnim, eAnim_SingleArms))
```
3UR: These originally work fine.

Issue #871, opened 2026-03-07
[Suggestion]drop whole stack


rhyscw:
### Describe the solution you'd like

could you add the ability to crouch then drop to drop a whole stack  

### Is your suggestion related to a problem?

its not related to a problem just nice to have 

### Additional context

it needs to be like newer versions of minecraft where you can drop a whole stack if crouching whilst dropping

PR #1159, opened 2026-03-11
 Fix: Dropping items outside of inventory menus

Adem-Kurt:
## Description
This PR fixes two separate bugs that prevented players from dropping items with clicking outside the inventory in menus.

### Previous Behavior
- In the Creative inventory, right-clicking outside the menu bounds while holding a stack of items did not drop a single item. Instead, the right-click input was intercepted to clear the hotbar.
- In standard container menus (Survival Inventory, Chests, Furnaces, etc.), clicking outside the menu bounds with an item held did nothing. The item would remain attached to the cursor.

### Root Cause
1. **Creative Menu:** In `IUIScene_CreativeMenu.cpp`, `handleValidKeyPress` had a check that captured any right-click (`buttonNum == 1`) and triggered the hotbar clearing function, terminating the input flow before the outside drop logic could process it.
2. **Standard Menus:** In `AbstractContainerMenu.cpp`, the `clicked` function had an early bounds check (`if (slotIndex < 0 || slotIndex >= slots.size()) return nullptr;`). Because clicking outside the UI sends a slot ID of `SLOT_CLICKED_OUTSIDE` (-999), this early check blocked the input, never allowing the drop code to execute.

### New Behavior
- Players can now drag items outside the Creative inventory and right-click to drop exactly one item per click without unexpectedly clearing their hotbar.
- Players can now drag items outside standard containers (chests, crafting tables, standard inventory) and drop items (full stacks or single items) properly onto the ground.

### Fix Implementation
- **Commit 1:** Added `!m_bPointerOutsideMenu` to the right-click interception logic in `IUIScene_CreativeMenu.cpp` so it only clears the hotbar if the pointer is actually inside the UI boundaries.
- **Commit 2:** Added an exception for `SLOT_CLICKED_OUTSIDE` in the initial boundary bounds check in `AbstractContainerMenu::clicked` so the `-999` ID can properly reach the subsequent drag-and-drop code blocks.

### AI Use Disclosure
I used AI as a search tool, find the code causing this bug.

## Related Issues
- https://github.com/smartcmd/MinecraftConsoles/issues/871

### DIFF
--- a/Minecraft.World/AbstractContainerMenu.cpp
+++ b/Minecraft.World/AbstractContainerMenu.cpp
@@ -157,7 +157,7 @@ shared_ptr<ItemInstance> AbstractContainerMenu::clicked(int slotIndex, int butto
 	shared_ptr<ItemInstance> clickedEntity = nullptr;
 	shared_ptr<Inventory> inventory = player->inventory;
 
-	if (slotIndex < 0 || slotIndex >= (int)slots.size())
+	if ((slotIndex < 0 || slotIndex >= (int)slots.size()) && slotIndex != SLOT_CLICKED_OUTSIDE)
 		return nullptr;
 
 	if (clickType == CLICK_QUICK_CRAFT)


jvnpr:
Given this issue only occurs in the creative menu and not the inventory, how is it currently implemented in the inventory? Does the check only happen within the creative menu?

PR #1169, opened 2026-03-12
 Fixed DLC files not saving correctly + new 16x16 region file support

MarvelcoCode:
## Description
New worlds now use 16×16 chunk region files while existing worlds (made before this commit) remain on 32×32.

### Previous Behavior
All worlds used 32×32 chunk region files, which broke saved DLC's and new tut worlds

### Root Cause
it always used the 32x32 system and no support for 16x16 (which is the newer system)

### New Behavior
New worlds and DLC's saves use 16×16 region files. Existing saves (made before this commit)  automatically stay on 32×32 to avoid corrupting them.

### Fix Implementation
A marker (region_format_16) file is stamped into the save header at creation time for both empty newly created worlds and DLC's. RegionFileCache::_getRegionFile, _getChunkDataInputStream, and _getChunkDataOutputStream all check for this marker to decide between >>4 (16×16) and >>5 (32×32) region coordinates and chunk local offsets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,15 @@
-# MinecraftConsoles
-
-[![Discord](https://img.shields.io/badge/Discord-Join%20Server-5865F2?logo=discord&logoColor=white)](https://discord.gg/jrum7HhegA)
-
-![Tutorial World](.github/TutorialWorld.png)
+# BetterMinecraftLCE
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


MarvelcoCode:
basicly converting misinterprets region to chunk coordinate mapping and its impossible to fix without manaully changeing and even then you will still have rendering issue and it will lag alot

MarvelcoCode:
> @MarvelcoCode hey is it alright if I use your code to fix the same issues on my fork?

Yea ofc

PR #1173, opened 2026-03-12
FIX: Item Floating

MozzarellaRat:
Items now float in water.

### Description, Previous Behavior, Changes
Item Entities did not float in water. This PR adds an upwards velocity of 0.005f when in water.

### Root Cause
It was not implemented.

### New Behavior
Item Entities now float in water

### Fix Implementation
```
int tile = level->getTile(x, y, z);
if (tile == Tile::calmWater_Id) yd += 0.045f;
```
image


sylvessa:
items not floating is a intended feature, the floating was a thing apart of aquatic update on LCE, we will not be backporting stuff after TU24 for now.

PR #1183, opened 2026-03-12
Slim Skins

ItzSonicFaner:
[https://www.youtube.com/watch?v=xQfvLOPQsFs](https://www.youtube.com/watch?v=xQfvLOPQsFs)

## Description
This patch added the character Alex with slim arms that can be used on other future skins.

## Changes
1. Alex skin
2. Slim arms for Alex skin

### Previous Behavior
Before this change, when adding the Alex skin, the texture on the hands was displayed incorrectly.

### Root Cause
When I saw this problem, I decided to edit the code to add the ability to add such skins without big problems.

### New Behavior
In the skin selection menu, the game checks if the new skin from the skin list == a specific skin from the list, then the slim model is used instead of the default model. Roughly the same thing happens in the game itself, only the check for this skin is a little different, but everything works perfectly fine under my supervision

### Fix Implementation
The skin itself was added to the scripts: Textures.cpp, Textures.h, Definitions.h, UIScene_SkinSelectMenu.cpp, and Player.cpp. In HumanoidModel.h and HumanoidModel.cpp, I added body creation with a slim arms option, where switching a boolean would create either slim or non-slim arms. A skin check was added to UIControl_PlayerSkinPreview.cpp so that if the skin is, for example, the eighth (Alex skin), the arms should be slim. A slim model option was added to EntityRenderer.h and EntityRenderer.cpp, which will be empty if mobs are spawned instead of a player. PlayerRenderer.h and PlayerRenderer.cpp contain similar functionality as UIControl_PlayerSkinPreview.cpp, but for the first-person hand and similar features. LivingEntityRenderer.cpp contains new functionality for rendering a player model in a third- or second-person game, with or without slim arms. The functionality works as follows. PlayerRenderer.cpp in LivingEntityRenderer.cpp starts creating and rendering the regular and slim models, which are saved in the EntityRenderer.h and EntityRenderer.cpp scripts, but these models are used in the skin menu display script (UIControl_PlayerSkinPreview.cpp).

### AI Use Disclosure
I'll be honest, I tried using AI, but it didn't give me any good solutions. Ultimately, I found solutions by searching scripts for clues that would allow me to come up with and write logic in a script to implement slim models.

Edited: 64x64 skins support was added thanks to @Langtanium.


Langtanium:
I added support for 64x64 classic and slim skins for DLC Skin Packs. I will also be adding the 4th Birthday Skin pack. I will be adding my changes to your repository @ItzSonicFaner. 
Minecraft 3_13_2026 12_05_30 PM

sestain:
64x64 skins don't work anymore:
image

Langtanium:
@ItzSonicFaner, the proper way to implement the 64x64 skins from DLCs is to use the "anim" value stored in "m_uiAnimOverrideBitmask" to set the skin. I added the functionality back, but made sure to implement it better. I also renamed the second layer geometry to use its correct name (body_layer to jacket, arm0_layer to sleeve0, arm1_layer to sleeve1, leg0_layer to pants0, and leg1_layer to pants1) (the names for the parts have been derived from DLC skins) and added the code to hide them if certain bits are present in the skin's "anim" value. There is an issue where the second layer geometry does not follow the parts they are suppose to follow. Also when swapping the original model with the slim/classic one causes extra geometry in the DLC skins to not render, 64x32 DLC skins render normally. I do not know if you have already make a fix, but I have sent a Pull Request.

Langtanium:
@ItzSonicFaner, 64x32 slim skins do not exist and the code you have added for it never gets used.

codeHusky:
Closed in favor of #1453 's implementation of this for now

PR #1195, opened 2026-03-12
Stained Glass Survival Integration & Crafting UI Fix

LordCambionGaming:
## Description
This PR restores survival access to stained glass and stained glass panes by implementing a custom crafting logic and fixing a UI constraint in the crafting menu.

### Previous Behavior
Stained glass and panes were only accessible in Creative mode. Their recipes in "ClothDyeRecipes.cpp" were disabled via preprocessor directives. Adding new recipes to the crafting grid would cause layout issues, creating new columns that overlapped or deleted existing columns (like the Rockets column).

### Root Cause
Stained glass recipes were wrapped in an "#if 0" block in "ClothDyeRecipes.cpp", the vertical slot size for the crafting grid was hardcoded to 17, which limited the number of displayable categories and "thinGlass" (glass panes) lacked the same base item type and material definitions as stained variants, causing them to be categorized separately.

### New Behavior
Stained glass and panes are now fully craftable in Survival. The crafting UI remains stable and organized, with all glass variants grouped into the same column:
immagine
### Fix Implementation
I Removed the disabled code from "ClothDyeRecipes.cpp" and implemented the new crafting logic within "StructureRecipes.cpp " for  a better organization.
I also increased the vertical slot height in the crafting UI from 17 to 99. This allows for more recipes without breaking the column layout.
Then I updated "thinGlass" using "setBaseItemTypeAndMaterial" to match the stained glass variants. This ensures they are grouped correctly in the same crafting column.

### AI Use Disclosure
I didn't use artificial intelligence; I did my own research and modified and read the code myself, without any outside help.
I apologize if I made any spelling mistakes, but English isn't my first language.

## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]


Bakoubak:
I would have removed the #if 0 instead of replacing it with 1

PR #1234, opened 2026-03-14
Fix input for Russian keyboard layout

St4lker0k765:
## Description

Add proper support for Russian letters input

image

### Previous Behavior

Previously after attempting to write any Russian text in chat it would just show off as some gibberish, even though Russian letters are supported by Mojangles font, and there are no problems if you try to copy and paste Russian text into game chat.

### Root Cause

Window message `WM_CHAR` can only contain 256 possible letters from Windows-1252 encoding, and it's fine for English and most of the European languages, but not for Cyrillic languages like Russian (it uses Windows-1251 encoding).

### New Behavior

Now characters convert to Windows-1251 if Russian keyboard layout is active.

### Fix Implementation

Added `TranslateSymbolUsingCP1251` function to `WM_CHAR` call which activates if active keyboard layout is Russian (`00000419`).

### AI Use Disclosure

No AI was used to create this commit.


codeHusky:
Are there no ways to fix this without hardcoding the table?

PR #1245, opened 2026-03-14
Fix: Tooltips Work Properly in Splitscreen

tylerreese543-lab:
Fixed tooltip changes

## Description

Fixed the controller tooltips to acknowledge individual pads instead of the primary pad for flagging rendering.

## Changes

Code now has flags set depending on which player is being rendered.
### Previous Behavior

Tooltips would only render based on the whether the first player's menu status (they shouldn't even render in the menu like that anyway i dont think)
Screenshot 2026-03-14 142422
Screenshot 2026-03-14 142432
Screenshot 2026-03-14 142623

### Root Cause

Return signal before rendering calculations; if the tooltips weren't meant to render for player ONE, they didn't render for any of the players.

### New Behavior

Now it queries the settings of each individual pad for whether to render.

image

(Player one in this example has tooltips turned off)
### Fix Implementation

Added two booleans (renderTooltipsOK, menuOKoverride) and one int (plr_pad)
plr_pad is set from a switch statement that gets the pad based on what enumerators were passed into the UI render function.
For example a top-splitscreen player would get pad zero and a bottom-left player would get pad two. I'm not sure if the way I retrieved pads is the most reliable, but it should be a really easy fix if requested.

Anyways that plr_pad is used to query state and settings which flags renderTooltipsOK
an if-return statement is called for renderTooltipsOK so that tooltips are only rendered when necessary

menuOKoverride is initially false, but can be set to true if the level or player does not exist (i.e. the game state is in the main menus)

This way tooltips will render regardless

### AI Use Disclosure

i aint no CLANKER
## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]

### DIFF
--- a/Minecraft.Client/Common/UI/UIComponent_Tooltips.cpp
+++ b/Minecraft.Client/Common/UI/UIComponent_Tooltips.cpp
@@ -212,8 +212,7 @@ void UIComponent_Tooltips::render(S32 width, S32 height, C4JRender::eViewportTyp
     }
 
 
-	if(m_bSplitscreen)
-	|
+	if(m_bSplitscreen) {
 		//Individiualize the flags for each viewport :)
 		S32 xPos = 0;
 		S32 yPos = 0;


tylerreese543-lab:
Sorry. The predictive text in GitHub Copilot replaced a open { with a |, causing the code to error and not build correctly. I had to re-commit to fix.

PR #1247, opened 2026-03-14
Java texture packs

Soggy-Pancake:
## Description
Adds java texture pack support

## Changes
Loads texture packs from zip files.

### Previous Behavior
DLC pack files only

### New Behavior
Texture packs are found on game start and are added to the list of available packs.
image
image

### Fix Implementation
- Added bit7z library with 7z.dll for zip file loading
- Implemented ZipFile, ZipEntry, and FileTexturePack classes
- Updated all tile and item names to match current versions
- Auto atlas generation using matched names to find textures
- Most atlas icon positions are now tied to the item/tile instance name to allow easier name changes in the future
 - Not all were able to be updated as they have multiple icon variants

### AI Use Disclosure
Some scanning to understand parts of the codebase and as interactive documentation of cpp.

## Current issues
- No folder loading yet
- Packs <16px are busted by animated textures
- Partial blocks like end portal frames and beds break with smaller packs as well due to rounding.


Soggy-Pancake:
Folders now load as well as animated textures that aren't animated by default are now cropped to the first frame. I'm unsure of how to fix blocks with models occasionally breaking with packs <16px. It is completely visual though and doesn't impact most packs people would use.

PR #1265, opened 2026-03-15
fix: removed headlesss server logic in favor of the new server exe.

mattsumi:
## Description
remove the headless server functionality in the client binary (using the old -server flag) in favor of the standalone dedicated server executable.

## Changes
removed [HeadlessServerCtrlHandler], [SetupHeadlessServerConsole], [HeadlessServerConsoleThreadProc], and [RunHeadlessServer] from Windows64_Minecraft.cpp

removed -server flag parsing and [serverMode] field from Win64LaunchOptions

removed dedicated server global initialization ([g_Win64DedicatedServer], [g_Win64DedicatedServerPort], [g_Win64DedicatedServerBindIP]) from client argument parsing

simplified -ip and -port args to only handle client multiplayer join

XUID resolution now runs unconditionally on startup

window is always shown (no more SW_HIDE since -server no longer exists.)

### Previous Behavior
the client executable doubles as a headless server when launched with -server, duplicating functionality that the dedicated server exe now provides.

### Root Cause
The headless server was the original server implementation before minecraft.server was introduced. it's completely unnecessary now that the full dedicated server is implemented.

### New Behavior
The client is client-only. All server functionality is handled by minecraft.server and the exe it compiles.

### Fix Implementation
removed all headless server code paths from Windows64_Minecraft.cpp.
the g_Win64DedicatedServer globals and conditional code in shared files (MinecraftServer.cpp, PlayerList.cpp, WinsockNetLayer.cpp, etc.) are preserved as they're still used when building minecraft.server. 

### AI Use Disclosure
No AI was used for this PR.

### DIFF
--- a/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
+++ b/Minecraft.Client/Windows64/Windows64_Minecraft.cpp
@@ -121,7 +121,6 @@ static WINDOWPLACEMENT g_wpPrev = { sizeof(g_wpPrev) };
 struct Win64LaunchOptions
 {
 	int screenMode;
-	bool serverMode;
 	bool fullscreen;
 };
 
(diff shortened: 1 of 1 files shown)


mattsumi:
> The impact seems smaller than I expected. I thought there would be more changes around commands, `server.properties`, saving, and other areas.

Yeah. It's not a massive change. Lots of the functions for the headless also came over for the actual deadicated server.

PR #1279, opened 2026-03-15
organized keybinds a lil bit

retuci0:
## Description
organized all keyboard binds to be in KeyboardMouseInput for easier customizability

### Previous Behavior
keybinds were hard coded in Windows64_Minecraft.cpp

### Root Cause
easier customizability

### New Behavior
now all keybinds are defined in one place: KeyboardMouseInput

> no ai was used, obv


3UR:
looks good to me

Issue #783, opened 2026-03-07
[Bug] Sound effects are really quiet


HunterBoy344:
### Bug description

Certain sounds, such as a chest opening, are now extremely quiet.

### To Reproduce

1. Open a chest
2. Observe that the sound is much quieter than it should be

### Expected behavior

Sounds should play at the correct volume.

### Videos

https://github.com/user-attachments/assets/fe38df0c-4446-4924-9b20-4255ad0371c0

### Version

13960a9

### Build Type

- [x] Release
- [ ] Debug

### Additional context

This was not an issue with a nightly build from this morning. I am suspicious of 55be6447e33e27cc8820b3fa7d6a8decd147cd2e...

Criseda:
Can confirm this is happening to me too

adbr3:
this is happening to me too

PR #1282, opened 2026-03-15
Fix audio volume rolloff and some missing/incorrect sounds

### LINKED ISSUE #1086: [Bug] All sounds are super quiet except music and menu
### Bug description

So basically, when playing the game, music and menu sound effects are good volume, but literally everything else is super quiet in comparison.

### To Reproduce

Just anytime you open the game.

### Expected behavior

For all sounds to be equally loud

### Screenshots

_No response_

### Videos

_No response_

### Version

latest nightly

### Build Type

- [x] Release
- [ ] Debug

### Additional context

_No response_

TorutheRedFox:
## Description
Fixes issues with sound volume, incorrect explosion sound playing and portal sounds not playing at all.

### Previous Behavior
People complained of sounds being too quiet or not playing at all

### Root Cause
Sound rolloff was set up incorrectly by using the wrong algorithm (miniaudio's default isn't linear while Minecraft expects linear), and in the case of jukeboxes, the 3d parameters weren't set up at all.

### New Behavior
Sound volumes have been corrected by trying to match early Java Edition's use of Paul Lamb's SoundSystem and the correct explosion sounds now play. All portal sounds now play as they should too.

### Fix Implementation
The correct algorithm is explicitly specified when playing sounds using miniaudio, and 3D audio parameters were tweaked to ensure the correct rolloff and volumes (rolloff increased to 1.0 from 0,5, max volume reduced to 1.0 from 1.5 as compensation is no longer necessary, etc.). Explosion sounds were fixed by removing explode.ogg and portal sounds were fixed by making SoundEngine::PlayUI also search the base sound folder (Windows64Media\Sound\Minecraft) if it can't find the requested sound in the UI subfolder. The volume of PlayUI was also corrected according to Java Edition's behavior (multiply the volume by 0.25 after clamping it), making the portal sounds that use it not ear-piercingly loud.

### Notes
Chest sounds are still quiet, though that was an intentional change on 4J's part because someone deemed it to be ["much louder than other sounds"](https://github.com/smartcmd/MinecraftConsoles/blob/e9f5b4b6f0bff50691acae160434c69c4a89231e/Minecraft.World/ChestTileEntity.cpp#L308).

### AI Use Disclosure
No AI has been used to make this PR.

## Related Issues
- Fixes #1086 & #783 (properly)


TorutheRedFox:
Demonstration of fixed audio
https://github.com/user-attachments/assets/af8e2433-7a13-4538-b3cd-8e05c80f874a

sylvessa:
seems sane, will test

sylvessa:
https://github.com/user-attachments/assets/475638c0-54cc-4d53-9e70-ea4c2049018f

any reason why it plays the title sounds a bit messed up?

TorutheRedFox:
i haven't changed anything except add a fallback path for ui audio and change parameters for audio rolloff

the difference could just be perceptual as the ui audio is lower in volume

codeHusky:
The UI sounds are definitely being played back in a way that makes them sound doubled instead of stacked. If there's an underlying bug thats causing that (which this PR only makes noticeable) then it'll still need to be fixed before it's merged

PR #1285, opened 2026-03-16
Fix game crashing if DLC has XMLVERSION paramater

sestain:
## Description
Fixes a crash that was caused by XML Version paramater being at the spot which would be uiFileCount usually (This means the DLC has advanced BOX parameters, at least to my knowledge that's what it means).

### Previous Behavior
Game would've crashed with DLCs which were too new.

### Root Cause
If a DLC (.pck) had parameter "XMLVERSION" it would cause uiFileCount to be the XML version number (so a low number like 1-3).
This caused the game to crash.

### New Behavior
Now it skips over the version number if it exists, and game doesn't crash.

### Fix Implementation
Added a bool before parameters for-loop, check for parameter named "XMLVERSION", and after for-loop we use the bool to see if we skip 4 bytes or not so uiFileCount is correct.

### AI Use Disclosure
No AI was used.

## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]


sestain:
> Shouldnt we be using the version of the dlc file not if this flag exists?

Not possible, since .PCK version goes only from 0 to 3, and "XMLVERSION" is just a string inside the parameters which later version of the game checks for to see if it needs to offset uiFileOffset.
DLCs that has that parameters are .PCK version 3 also (afaik).

PR #1312, opened 2026-03-17
added zoom with F

retuci0:
## Description
zooms in the camera while holding F

### Previous Behavior
no zoom

### Root Cause
zoom is convenient

### New Behavior
you can now zoom the camera by a factor of 5 by holding F

### Fix Implementation
uses GameRenderer::zoomRegion() (fixed, btw) to zoom in, and GameRenderer::unZoomRegion() to zoom back out.

### AI Use Disclosure
no ai was used.


antgame11:
too non vanilla imo

PR #1325, opened 2026-03-18
Increase limits for boats, minecarts, fireballs, and height

Jazzitch:
## Description
Increases the world height to 512, and increases entity spawn limits.

test

### Previous Behavior
The world height was 256. This limited the height players could built to. Also, only a few entities were allowed. This was critical for the old ~512 mb limit on old consoles, but modern computers don't need these limits.

### Root Cause
The hard-coded limit was 256 which did not allow tall builds such as a castle, especially on mountains. Boats, minecarts, etc could not exceed a certain amount (ex. 40), which could hinder builds such as roller coasters.

### New Behavior
Players can build up to 512 blocks up on the y-axis. Boats, minecarts, and projectiles can exceed the old limits.

### Fix Implementation
I increased the old build limit by x2, and the old entity limits by x10.

### AI Use Disclosure
No AI was used.


mattsumi:
Thank you for contributing!

PR #1350, opened 2026-03-19
Fix: The buttons in the some menus do not make a clicking sound

Adem-Kurt:
## Description
The buttons in the Help and Options menu, Debug menu, Death menu and Pause menu make a clicking sound after this update.

### Previous Behavior
The buttons in the Help and Options menu, Debug menu, Death menu and Pause menu do not play a clicking sound.

### Root Cause
While the `handlePress` function in other menus (such as Settings and Settings Options) have code that plays a click sound, but that code is not in the Help and Options menu, the Debug menu, the Death menu, or the Pause menu 

### New Behavior
The buttons in the Help and Options menu, Debug menu, Death menu and Pause menu play a clicking sound.

### Fix Implementation
I added the line `ui.PlayUISFX(eSFX_Press);` at the beginning of the `handlePress` functions in the menu files (as in `UIScene_SettingsMenu.cpp`, `UIScene_SettingsOptionsMenu.cpp`, etc.).

### AI Use Disclosure
None

## Related Issues
- None

### DIFF
--- a/Minecraft.Client/Common/UI/UIScene_DeathMenu.cpp
+++ b/Minecraft.Client/Common/UI/UIScene_DeathMenu.cpp
@@ -81,6 +81,8 @@ void UIScene_DeathMenu::handleInput(int iPad, int key, bool repeat, bool pressed
 
 void UIScene_DeathMenu::handlePress(F64 controlId, F64 childId)
 {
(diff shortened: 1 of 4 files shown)


Adem-Kurt:
I'm closing this because I want to make better solution to this problem.

PR #1384, opened 2026-03-22
Allow displaying item lore

sylvessa:
## Description
This uncomments the code that allowed for displaying lore under item names.

### Previous Behavior
Lore for items would not be shown under item names
image

### Root Cause
4J made Item lore not be shown due to it not being used anywhere really, but kept the code in place just commented out.

### New Behavior
Item lore now shows under item names
image

### Fix Implementation
Uncommented the code that was commented out for Lore displaying

### AI Use Disclosure
no ai


UniversePM:
ai use disclosure is wrong, you used ai to uncomment the already existing code

you're banned!!

PR #1407, opened 2026-03-25
Use proton and wayland for the dedicated server

Twig6943:
## Description
Stop using wine and xvfb for the dedicated server build

## Changes
Use wlheadless with cage for the display and umu-launcher for proton. Just like [this](https://github.com/silkker/MARNE/blob/main/Containerfile) [example](https://marne.io/servers)

### Previous Behavior
Used debian12, wine & xvfb

### Root Cause
Wayland is the way forward?

### AI Use Disclosure

No.


Twig6943:
Is it fine if I create a non-root user for umu?

image

PR #1409, opened 2026-03-25
Update README.md

maydoxx:
## Description

Because there's no APK in releases and in the code there's no Kotlin or anything that should be used to compile for Android
## Changes
Just changed the README
## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]


README.md
```diff
@@ -19,7 +19,7 @@ If you're looking for Dedicated Server software, download its [Nightly Build her
 
 - **Windows**: Supported for building and running the project
 - **macOS / Linux**: The Windows nightly build will run through Wine or CrossOver based on community reports, but this is unofficial and not currently tested by the maintainers when pushing updates
-- **Android**: The Windows nightly build does run but has stability / frametime pacing issues frequently reported
+- **Android**: No current support
```
fancythedeveloper: It would make a lot more sense to rewrite this line to be similar to the one above, stating that it does run under apps like GameHub/Winulator, though has performance issues, and is still unofficial regardless.

codeHusky:
This line will be updated, thanks for the feedback. We will not be using your suggestion, tho

PR #1454, opened 2026-04-01
Minor docker/container fixes and edits

SnenxyTengoku:
## Description

Just a few (no, really) simple line edits and changes for the containerized configuration and setup of the dedicated server, as these were just a few minor fixes I decided to just do it myself.

(Worth noting that I believe that the MCLCE dedicated server is unable to read `server.properties`, although making *that* part work is outside of my available knowledge)

## Changes

- Do not hardcode the IP address and port, add default environment variables in compose files to reflect the change/suggest users to allow changing it
- Point remote image of the `ghcr` compose file to the right location (`ghcr.io/kuwacom` -> `ghcr.io/smartcmd`)
- Use more generic `Etc/UTC` timezone by default
- Copy the default display resolution envvar from the `ghcr` compose file to the manual build one.

### Previous Behavior

- The IP address and port was permanently set by the entrypoint script for the container, which meant that containerized users were completely unable to set their own IP address and port if they wished to. This is an issue for users who already run a Java edition server on the same machine, or for those who simply wished to use another port that isn't the default/well known one.
- The `ghcr` compose file would pull in the image from `kuwacom`'s account. Was most likely just overlooked during the merge of the dedicated server stuff.
- The default timezone was set to `Asia/Tokyo`.

### Root Cause

- The entrypoint script used by the MCLCE container set `SERVER_PORT` and `SERVER_BIND_IP` to `25565` and `0.0.0.0` respectively. It was not possible to change this at all unless one were to manually set the entrypoint/execution command docker used while starting the container, which isn't exactly clean.
- The `ghcr` compose file had the image value set to `ghcr.io/kuwacom/minecraft-lce-dedicated-server:nightly`. Definitely just an overlooked leftover from the initial dedicated server pull request #498.
- The `TZ` environment variable was initially set to `${TZ:-Asia/Tokyo}`. 

### New Behavior

- Container users can now set the server IP address and server port by setting the `SERVER_BIND_IP` and `SERVER_PORT` environment variables inside of the compose file/however it gets run.
- The `ghcr` compose file now pulls in from `ghcr.io/smartcmd/minecraft-lce-dedicated-server:nightly`.
- Use `Etc/UTC` as the new default timezone.
- Use the set value of `720x1280x16` from the `ghcr` compose file in the manual build compose file. Probably an unnecessary change, but uniformity + saner value imo.

### Fix Implementation

- Modify `./docker/dedicated-server/entrypoint.sh` to give the `SERVER_PORT` and `SERVER_BIND_IP` values a user changeable value, with a default value of `25565` and `0.0.0.0` if not set by the user.
- Modify `image` line to point to the correct `smartcmd` image location.
- Modify TZ -> `${TZ:-Etc/UTC}`
- Copy `XVFB_SCREEN`'s value from `ghcr` to manual build (`64x64x16` -> `720x1280x16`) 

### AI Use Disclosure

One should not need AI to be able to make these changes tbh


sylvessa:
these r good changes. THank u for this

PR #1468, opened 2026-04-04
Migrate window class registration and window creation to Unicode APIs

rucheykov:
## Description
Migrate window class registration and window creation from ANSI (RegisterClassA / CreateWindowA) to Unicode (RegisterClassW / CreateWindowW) APIs. This ensures proper handling of non-ASCII characters in text input.
## Previous Behavior
Window class and windows were created using ANSI APIs.
WM_CHAR messages delivered characters encoded in the system code page, leading to incorrect display of non-ASCII text (e.g., Cyrillic, accented characters).
## Root Cause
The use of ANSI APIs meant that character data was limited to the system code page. Non-ASCII input could not be represented correctly, resulting in garbage characters in text fields.
## New Behavior
Window class registration and window creation now use Unicode APIs.
WM_CHAR messages now provide correct UTF-16 codepoints.
Non-ASCII characters are properly entered and displayed in text fields.
## Fix Implementation
Replaced RegisterClassA with RegisterClassW and updated WNDCLASS/WNDCLASSEX structures to use wide-character strings.
Replaced CreateWindowA with CreateWindowW and ensured all string parameters are UTF-16 (L"...") literals.

**Note:** This will NOT solve the problem of entering Chinese and other characters; they simply aren't in the font. For that, we need to implement a fallback to GNU Unifont.
**Note2:** This is a remake of the request https://github.com/smartcmd/MinecraftConsoles/pull/950


Minecraft.Client/Windows64/Windows64_Minecraft.cpp
```diff
@@ -675,23 +675,23 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
...
 	wcex.hInstance		= hInstance;
-	wcex.hIcon			= LoadIcon(hInstance, "Minecraft");
+	wcex.hIcon			= LoadIconW(hInstance, L"Minecraft");
 	wcex.hCursor		= LoadCursor(nullptr, IDC_ARROW);
 	wcex.hbrBackground	= (HBRUSH)(COLOR_WINDOW+1);
-	wcex.lpszMenuName	= "Minecraft";
-	wcex.lpszClassName	= "MinecraftClass";
+	wcex.lpszMenuName	= L"Minecraft";
+	wcex.lpszClassName	= L"MinecraftClass";
 	wcex.hIconSm		= LoadIcon(wcex.hInstance, MAKEINTRESOURCE(IDI_MINECRAFTWINDOWS));
```
sylvessa: wouldnt we want to make this LoadIconW too?

Issue #1440, opened 2026-03-29
[Bug] Server connection lost occasionally when trying to autosave


zigaloid22:
### Bug description

On the Legacy Fortress server, using the latest commit, we experience server connection errors where it boots us from the game and prevents us from connecting. When we try to rejoin, the connecting loading bar freezes and doesn't move at all and I have to close and reopen my game. The server has to be reset to let us join again because the game thinks we are still in. 
This is what the person running the server said: "Yeah, it seemed to happen during autosave, and the server console outputs something about critical region. I assume that a thread got stuck in the critical region during autosave and locks up the other threads."

### To Reproduce

1. Have a dedicated server on the latest commit
2. Have autosave activated
3. Autosave will occur and might disconnect players (Not every time)
4. Try and rejoin
5. Connecting bar not loading, basically frozen

### Expected behavior

The autosave to work and not kick us and prevent us from connecting.

### Screenshots

Well this is what it looked like in the server console output.

<img width="967" height="454" alt="Image" src="https://github.com/user-attachments/assets/f77983f5-f63d-4f8d-91e9-9d6b5b673748" />

### Version

38d58f2d

### Build Type

- [x] Release
- [ ] Debug

PR #1473, opened 2026-04-05
optimization: async dedicated server autosaving

sylvessa:
## Description
makes autosave async on the dedicated server so it doesnt block the tick thread during compression

### Previous Behavior
calling `Flush()` on the server would block the entire tick thread while it compressed. No idea why this is ran on main thread especially in a server, I'm pretty suire this was changed later on. for large worlds this caused noticeable lag and made the server practically freeze during autosaves.

from 1 autosave on a 350mb world ran on the dedicated server (latest nightly)

https://github.com/user-attachments/assets/480cba7a-5e6a-4941-85dc-22b99b315902

### Root Cause
`Flush()` did all compression work synchronously. the tick thread had to sit and wait for zlib to finish compressing the full save buffer before it could continue. this caused the server to freeze for a few seconds depending on the world size and hardware.

### New Behavior
`Flush()` snapshots `pvSaveMem` into a heap buffer under the critical section, releases the lock immediately, then hands the snapshot off to a seperate thread that does all the compression. once the bacjground thread finishes, the result gets picked up the main thread next tick and commits it to `StorageManager`.

also when server stops it waits until this all does its thing, i had ran into an issue where it would just do the save in a seperate thread then just stop without actually writing, so it waits until thats done

With this, theres no noticable lag when autosaving, users can still play chat etc while autosave occurs

https://github.com/user-attachments/assets/accaaa46-9b7a-4122-a16b-23ed78db5481

3 autosaves happened in this video, on a 350mb world ran on the dedicated server with the async autosave

### Fix Implementation
ConsoleSaveFileOriginal now has two new funcs, flushPendingBackgroundSave which gets called bny the tickcoresystems in servermain.cpp, and hasPendingBackgroundSave for checks by servermain.cpp before closing down the server fully

`Flush()` now gathers the snapshot which contains all the save data and etc, and then writes to disk etc in its own thread, Result sits until `flushPendingBackgroundSave` gets picked up by the tickcoresystems and then actually commits to `StorageManager`

StorageManager is Unfortunately not thread safe (oops) so the actual commit has to stay on the main thread.

flush() onlt runs in a seperate thread IF set to dedicated server btw. I made sure not to touch the save stuff when ran as client lol.

### AI Use Disclosure
no 

## Related Issues
- PROBABLY Related to #1440


sylvessa:
merging since mattsumi has looked through it

PR #1474, opened 2026-04-05
feat. set intro scene background to white

Firebladedoge229:
## Description
fixes the intro swf on startup having a black pillarbox instead of a white one

## Changes
set the scenes background to white

### Previous Behavior
scene background was black

### Root Cause
not setting a specific background color

### New Behavior
setting a specific background color when the intro scene is showing

### Fix Implementation
setting the background to white

image


sylvessa:
confirmed this works, will merge soon

Issue #900, opened 2026-03-08
[Suggestion] Whitelist for servers (and potential increase in the amount of hex bytes for the uid)


gizmogoat:
### Describe the solution you'd like

There are some current concerns with network exploits in the codebase. It would be nice if there's some file that let you include your friend's UIDs so only they can join. To make it harder to bruteforce, it might be worth doubling or tripling the size of the UID if possible.

### Is your suggestion related to a problem?

Private servers with friends are desirable, especially since there are likely many security vulnerabilities in the netcode that could lessen their impact in the meantime by hand-curating trusted individuals to join.

codeHusky:
I personally believe having a "server password" system would probably make more sense than increasing the size of UIDs, at least for over-the-network, zero-server whitelist auth. Good two factor system that verifies both the identifier for the save and a shared secret (literally a password) for connecting to the private server.

gizmogoat:
> I personally believe having a "server password" system would probably make more sense than increasing the size of UIDs, at least for over-the-network, zero-server whitelist auth. Good two factor system that verifies both the identifier for the save and a shared secret (literally a password) for connecting to the private server.

That works too yeh

Issue #613, opened 2026-03-05
Player Experience Improvement TODO!


codeHusky:
In no particular order...
- [ ] Add basic offline-mode profile system
  - [ ] 4JLibs offline mode profile in 4J_Profile
  - [ ] Make sure we're using player defined UID to allow username changes
- [ ] Make Multiplayer work better (or at all?)
  - [ ] Split dedicated server into separate EXE (don't use client .exe)
  - [ ] Add moderation commands back in
  - [ ] Implement chat for KBM as well as controller
  - [ ] Add non-moderation op commands back in
  - [ ] Remove all of the awful WINDOWS64 preprocessor hackaround blocks for invalid XUIDs and use them properly!
- [ ] Improve UX for server list (button to add servers, etc)
- [ ] Improve scroll behavior in server list and save list on KBM
- [ ] Improve movement when flying in KBM (very rigid/fixed angles)
- [ ] Bring in-game menus (inventory, creative menu, etc) as close to MC:JE KBM experience as possible 
  - [ ] while retaining controller support
- [ ] Add support for Steamworks networking / p2p / profiles (4JLibs 4J_Profile type thing)
- [ ] Add support for more controllers outside of just XInput / Xbox controllers

If you'd like to contribute changes that address anything in this list - please do! PRs are welcome and appreciated.

PR #1475, opened 2026-04-05
feat: Add login verification flow (aka: auth v3)

mattsumi:
## Description

1. Implements a new login packet exchange to handle verifying users during the "login" step, both in "online" and "offline" type authentication schemes
    - Kicks players who fail to authenticate
    - Fully configurable by the server (which "auth" method, including `offline`)
    - Defaults to `offline` which is the same as how servers work today
    - All "online" methods are opt-in and are not required by either the client (users) or servers
    - Clients must have matching logins to connect to authenticated servers.
2. Implements yggdrasil-compatible auth (ely.by, third party, MC: Java Edition)
3. Implements a no-op Offline "auth" that simply passes through the username/uid from the client (previous behavior)
4. Adds libcurl for HTTPS requests
5. Implements a full Java Edition oauth login routine (MSFT Oauth -> Xbox Game Services -> Mojang Session)
6. Adds a persistent profile manager for storing user logins / offline profiles
7. Switches from XUIDs to 128-bit UUIDs

note: (for now we are gonna keep the old xuid data, but push it out with a future patch.)

## Changes

see new behavior

### Previous Behavior

the original codebase was built for consoles where platform-level auth (xbl, psn, etc) handled identity. as we port it to windows, none of this exists, so we are faking this identity. as such, there is no HTTP client to talk to auth servers, no handshake protocol to negotiate authentication schemes, and no UUID system to give players a stable cross-platform identity. the [PreLoginPacket] name field could also overwrite an already-authenticated username which was a pretty obvious security hole once real auth was in place. big no no!!! (also msft store button was going unused.)

before

### Root Cause

the original codebase was built for consoles where platform-level auth (Xbox Live, PSN) handled identity. when the project got ported to Windows none of that infrastructure existed, so identity was basically faked. there was no HTTP client to talk to auth servers, no handshake protocol to negotiate authentication schemes, and no UUID system to give players a stable cross-platform identity. the [PreLoginPacket] name field could also overwrite an already-authenticated username which was a pretty obvious security hole once real auth was in place.

### New Behavior

players now authenticate through one of four methods. **msft** (oauth device code flow that opens a browser and polls for completion), **yggdrasil** (username/password exchange against any yggdrasil-compatible provider — ely.by ships in [yggdrasil.json], mojang is hardcoded as fallback), **keypair-offline** (local asymmetric keypair auth supporting rsa2048 and ed25519), or **offline** (trusts users are who they say they are). the server declares which auth scheme it requires via a new [auth-mode[ property ([session] or [offline]), and a full 11-stage handshake protocol ([ANNOUNCE_VERSION] through [AUTH_SUCCESS]/[AUTH_FAILURE]) negotiates and verifies identity before the player ever gets past the login screen.

~~(funnily enough, msft sees us as minecraft for nintendo switch when autodetected. not bedrock, the original.)~~

This is now fixed; We are now recognized as Java, as we should have been.

Capture

continuing:

profiles now exist in a binary "auth_profiles.dat" file with token refresh support for both Microsoft and Yggdrasil providers. profiles now track a [variation] field so the system knows which provider issued them. the in-game UI lets you add, remove, cycle through, and apply profiles. players get a [GameUUID] (v5 UUID derived from their identity under the [MinecraftConsoles] namespace) that gets serialized in [LoginPacket] and [AddPlayerPacket]. existing ban/whitelist entries with XUIDs automatically get backfilled with UUIDs on load, and player save files get migrated from XUID-named ".dat" files to UUID-named ones.

the auth module hierarchy was consolidated in v3: [OfflineAuthModule] and [KeypairOfflineAuthModule] got inlined into [AuthModule.h] (separate .cpp/.h files deleted), [SessionAuthModule] was generalized to handle any yggdrasil provider via configurable URLs loaded from [yggdrasil.json], and the old hardcoded ely.by-specific [startElyBy()] flow was replaced with a generic [startYggdrasil()] that takes provider name and auth URL. [ELYBY] profile type became [YGGDRASIL].

if auth fails the client gets a proper [eDisconnect_AuthFailed] disconnect reason with a localized error message instead of just getting silently dropped.

after1
after2

### Fix Implementation

the core of this is a new [HttpClient], [AuthModule] hierarchy (session/keypair-offline/offline — all in [AuthModule.h] now), [AuthPackets] (11-stage enum with key-value fields), and a [HandshakeManager] state machine that drives the whole protocol from both sides. [SessionAuthModule] reads provider configs from [yggdrasil.json] at runtime and does full server-side [hasJoined] verification over HTTPS.

[AuthScreen] handles the threaded OAuth/yggdrasil auth flows and profile persistence in a binary [auth_profiles.dat]. [UUID.h/cpp] implements [GameUUID] with a custom SHA-1 for v5 generation, deterministic XUID conversion, and a [MCCONSOLES_NAMESPACE_UUID] constant. on the server side, [PendingConnection] now runs [initAuth()] based on [auth-mode] from [ServerProperties], guards [PreLoginPacket] name overwrites, and disconnects with [eDisconnect_AuthFailed] on failure. [ClientConnection] mirrors this on the client side and syncs the authenticated name back into the platform globals. [DirectoryLevelStorage] handles XUID-to-UUID save file migration, and [BanManager]/[WhitelistManager] backfill UUIDs on load. [UIScene_MainMenu] got the full profile management UI with next/use/add/remove.

### AI Use Disclosure

no ai was used to make this

## Related Issues
- Related to #613
- Related to #900
- Related to #1012 (?)


codeHusky:
Please refactor AuthModule, the auth implementations should not be in one large monolithic file. Separate them out.

Firebladedoge229:
i dont think we should do this genuinely
the whole point of lce, like what Cyber-Sec-DUBLIN, was to not have any form of microsoft affiliation / authentication in a way
this pr completely ruins that, destroying what we once had
and the fact that codehusky keeps calling anyone who has an opinion countering him a "child" doesnt necessarily make this any better
image

Firebladedoge229:
this also just gives microslop the opportunity to suspend anyone who uses the fork in the first place
in what world does this make sense?? we're quite literally just telling microslop atp: "HEY IM DOING SOMETHING ILLEGAL!!! PLEASE TAKE ACTION AGAINST MY ACCOUNT ACCORDINGLY!!!!"

funnyfella42:
yo u lowk just made a big ass arguement on the discord server how u feel?

WindowsDefen517:
I don't understand what's the point of this pull request, isn't the security checks enough?

Firebladedoge229:
hey man i think the 10 thumbs down emojis on this post should maybe  give you a sign that you should not merge this

Firebladedoge229:
>  > I have a question about this, would Microsoft actually know you're using the leaked LCE source code for this? I've seen tons of other software such as Prism Launcher allow you to login to you Minecraft account and there's been no issues that I've heard of, the login application doesn't need to specify that it's for logging into this and could be something simple like, "ConsoleLogin". Overall I don't think the login would need to send info to Microsoft telling them what you're logging into, correct me if I'm wrong.
> 
> No.
> 
> It just says "Minecraft Launcher."
> 
> No more details are provided.
> 
> 

they can just filter out authentication tickets for the phrase "Minecraft Launcher" with no additional details and take action accordingly

imo, merging this change would be a stupid + unnecessary decision

whitelists + non-public domains exist for a reason

codeHusky:
as has been noted already elsewhere, there are ways to appear as the original Minecraft launcher using this auth scheme. I'm pretty sure the current methods in here do as much but I haven't had time to review given all of the issue the project has had recently.

Either way, I'm very familiar with how the launcher login process works and it's not that hard to look like the real one. Dw.

WindowsDefen517:
Except maybe the 11 stage enforcement protocol for server, i still don't understand what's the point of this commit, except giving even more informations about us to microsoft. This commit is really not needed for now i think. And yes the fact that there are only 1 thumb up, and 10 thumbs down give a honest opinion about if we should include this merge or not.

fancythedeveloper:
> i still don't understand what's the point of this commit, except giving even more informations about us to microsoft.

Did you even READ the PR or are you just yappin' about something that's already been answered

dsgdhghgdhf:
Why do you need this?

dsgdhghgdhf:
> The irony of the original LCE leak coming about because somebody was angry at the direction Microslop was taking Minecraft only for us to be using literal Microsoft tech is crazzy

Yup... y'all dumb as fuck. I don't give a shit about it either being used for server-side or client side this whole authentication is bullshit and everyone knows it, people from Reddit, Discord, are telling you all this is a stupid idea and nobody listens. (yea they did, they even timed out ppl who spoke out & disagreeing, Worst PR management ive ever seen holy) This is the definition of the worst PR I have seen, in the same league as microslop, EA/Electronic Arts, Popcap, and Bethesda. You see the negatives and the positives and pull a completely self-inflicted lobotomy. People are already saying others are better and this (the company garbage sucker) is not a good option. This is technically still illegal code and trying to add authentication is still not going to work as a shield. At the end of the day, the foundational use of the stolen source code remains illegal if you are not the owner/manufacturer of the said code.
So this authentication is out the window immediately once it's questioned by the company.

And it's a shame for this to happen man... My brother even played this fork. I HELPED HIM SET IT UP.

Pushing more features is not a good idea, Pushing into Microsoft authentication could get the attention of Microsoft because of the outcry/uproar of adding authentication to a fork of the leaked code of LCE/ Minecraft Legacy Edition. And since there were already problems with this fork from the PR standpoint (something about a pred) this could make it to Microsoft unintentionally, which could make this fork get taken down.


And since this is running off of leaked source code and maybe from the points I have listed there is a clear warning from the community that forks, YouTube videos, or other content could be next if Microsoft continues its enforcement of taking down leaked code of their most profitable game, Minecraft.

And why the HELL would I want to give my password and email to a unofficial Minecraft client that even Microsoft or Mojang Studios Would NOT trust? I would give it to launchers that the community actually trusts. I know this is open source but no thanks. I would not like to not give my account password and email just so I can be verified for a leak of Minecraft. Even if it's only for verification to tell if you're not spamming multiple accounts for a server I still do not understand why do you gotta give them your Microsoft account details? Even if you don't own the game. I can understand why this is the easiest approach but please pick something else for god sakes.

And to add to my "concern" is that you are actively ignoring concerns even though the "contributors" have said "Keep this about PR related or "concern" about this topic" which has "surprisingly" Been ignored. Great job "people" at listening to concerns. You should really get a MeTaL.

WindowsDefen517:
> > i still don't understand what's the point of this commit, except giving even more informations about us to microsoft.
> 
> Did you even READ the PR or are you just yappin' about something that's already been answered

Maybe because it's not needed for this fork? I think the thumbs are already answering if we need this pr, 2 vs 12, i think it's enough to say that nobody want this pr.

codeHusky:
At this point we’re well aware of some of the community concerns over the Java Edition login flow. I’ve locked comments since it’s just getting spammy and repetitive at this point. If you’re looking to discuss this more please use the Discord going forward.

Edit: This PR has been closed and mattsumi has been removed as a contributor due to this being vibecoded

PR #1483, opened 2026-04-07
Add Chat Formatting Support For Servers

DrPerkyLegit:
## Description

adding chat formatting for servers

### Previous Behavior

chat wouldnt render html formatting
servers had no support for localization in messages
image_2026-04-06_223315531

### Root Cause

it just wasnt needed for p2p clients, chat was never added so they never bothered with the extra features

### New Behavior

the opposite of **Previous Behavior** except its sent from a server message, tested with fourkit as its designed for a server to send the message
image_2026-04-06_223403262
image_2026-04-06_223513047

### Fix Implementation

changes the chat labels to be html based so it can render html formatting
added support for multiple argument chat messages that also support formatting (this part is designed for the server, no default client access)
changed player messages to use &lt; and &gt; for <> because it breaks when using html labels

### AI Use Disclosure

i used ai to get the regex pattern cause i dont know how to make those honestly


DrPerkyLegit:
changed chat color to use the double s, client can send color, color doesnt effect chat message length (needs to be improved), html escaping the message to prevent any html parse exploits

DrPerkyLegit:
currently it only supports the basic color codes and the old translateables i showed, no bold / italics / corrupted text styling, something i plan on attempting to add, doesnt support multi argument messages anymore, clients can only send 100 characters (not including color codes), all messages cap at 255 regardless of color codes (i noticed a crash with large messages, client freezes and doesnt trigger the debugger)

DrPerkyLegit:
image
image

DrPerkyLegit:
prevented players from sending color but not the server as requested
image

codeHusky:
@DrPerkyLegit Could you resolve the merge conflict? Occurred after merging the chat scroll feature

DrPerkyLegit:
done

PR #1520, opened 2026-04-16
Chat Formatting Fix

### LINKED ISSUE #1515: [Bug] when you color 2 words in 1 message the message breaks
### Bug description

if i put § on 2 different words in one message i end up with this

### To Reproduce

in chat say anything with a §, then add another word with a § and theres the bug

### Expected behavior

i expected the 2 different words to be different colors

### Screenshots

Image

### Videos

_No response_

### Version

f78dec6e

### Build Type

- [x] Release
- [ ] Debug

### Additional context

2 § in 1 msg break it

DrPerkyLegit:
## Description

fixes a bug with the chat formatting
### Previous Behavior

color didnt apply after the first message
### Root Cause

if the string is too long it wraps around and breaks styling, doesnt respect visible message length

### New Behavior

color works correctly
image

### Fix Implementation

added functions to get the visible message length instead of raw message length
enforced html labels in UIScene_HUD::handleReload()
moved translateables from the chat handle function to the chat message format function
added the missing "§r" but with no styling we just use it as another "§f" option and make the color white

### AI Use Disclosure

no

## Related Issues
- Fixes https://github.com/MCLCE/MinecraftConsoles/issues/1515


Gatoel-capo:
yooo, cool, now i can make all things flashy

PR #1523, opened 2026-04-16
fix: Nether portal trigger and travel sounds not playing

### LINKED ISSUE #1516: [Bug] Nether Portal has no "traveling" sound
### Bug description

When traveling through a nether portal the file "trigger.ogg" does not play from the directory: LCEWindows64\Windows64Media\Sound\Minecraft\portal  . This is an issue in current version 2fba264c and seems to be an issue in many forks. This was not a bug in the official version of TU19.

### To Reproduce

1.Create a world
2.Go to a Nether Portal
3.Stand in the nether portal

### Expected behavior

To start "trigger.ogg" when in the portal and to start "travel.ogg" when exiting the portal.

### Screenshots

Image

### Videos

https://github.com/user-attachments/assets/1cd415dc-3c02-44f9-bcea-8778132c2fd6

### Version

2fba264c

### Build Type

- [x] Release
- [ ] Debug

### Additional context

This Bug seems to be in most builds/forks and has been in every build in the last 2 months following.

dtentiion:
## Description
Fixes nether portal sound effects on Windows 64-bit. The ominous `trigger.ogg` that should play when stepping into a portal and the `travel.ogg` whoosh that should play during dimension change were both silent. The only portal sound that worked was the ambient `portal.ogg` hum from nearby portal tiles.

https://github.com/user-attachments/assets/9fcf78ae-b178-4365-b803-4198ae73b728

Fixes #1516

### Previous Behavior
Standing in a nether portal made no sound at all. Getting teleported to the nether or back to the overworld made no sound either. The assets are present at `Windows64Media/Sound/Minecraft/portal/trigger.ogg` and `Windows64Media/Sound/Minecraft/portal/travel.ogg` but nothing would play from them. This was also reported to affect many forks.

### Root Cause
Two separate bugs on the same code path:

1. `SoundEngine::playUI` correctly selected the right subdirectory (`Minecraft/` for non-UI sounds, `Minecraft/UI/` for UI sounds) when looking up the name in the `wchSoundNames` / `wchUISoundNames` tables, but when building the actual file path it always hardcoded `Minecraft/UI/`:
   ```cpp
   sprintf_s(basePath, "Windows64Media/Sound/Minecraft/UI/%s", ConvertSoundPathToName(name));
   ```
   For `PORTAL_TRIGGER` and `PORTAL_TRAVEL` (which live in the non-UI table as `portal.trigger` / `portal.travel`) the resolved path was `Windows64Media/Sound/Minecraft/UI/portal/trigger.ogg` which does not exist, so `FileExists` returned false and the function bailed out before playing anything.

2. For nether/end portals, the dimension change is driven by `ServerPlayer::changeDimension` (via `Entity::tick` on the server), not `LocalPlayer::changeDimension`. `LocalPlayer::changeDimension` only fires from `TheEndPortal` tile, so its call to `playUI(PORTAL_TRAVEL)` never ran for regular portal travel. `ServerPlayer::changeDimension` had no equivalent call, so the travel whoosh never got triggered.

### New Behavior
- Stepping into a nether portal plays `trigger.ogg`
- Travelling to another dimension plays `travel.ogg` right before the dimension swap, followed by `trigger.ogg` a moment later as the player emerges inside the return portal
- Works for overworld -> nether, nether -> overworld, overworld -> end, and end -> overworld

### Fix Implementation
- **`SoundEngine.cpp` - `playUI`**: Store the chosen subdirectory (`Minecraft` or `Minecraft/UI`) in a local and use it when building `basePath`, so non-UI sounds resolve to `Windows64Media/Sound/Minecraft/<sound>` and UI sounds resolve to `Windows64Media/Sound/Minecraft/UI/<sound>`.
- **`ServerPlayer.cpp` - `changeDimension`**: Added a `playUI(eSoundType_PORTAL_TRAVEL)` call just before `PlayerList::toggleDimension` so the whoosh plays during the actual dimension swap. Guarded with a null check on `Minecraft::GetInstance()` and `soundEngine` so a headless dedicated server build won't try to play audio.

## AI Use Disclosure
No AI was used in the development of this pull request.


codeHusky:
these sounds are extremely loud but this does fix it. will merge

PR #1525, opened 2026-04-16
Remove breeding cap for animals

### LINKED ISSUE #1416: [Bug] This animal can't enter Love Mode.
### Bug description

whenever i try to breed any animals at all i get this message, this happened after i had a friend join and play with me for a while. i even tried cutting down almost all my animals in my animal farm, i still get this message. i had about 2x the amount of animals prior and never got this message.

### To Reproduce

make an animal farm, have a friend join and leave afterward, try to breed animals.

### Expected behavior

this message should not pop up after i kill half of my animals.

### Screenshots

_No response_

### Videos

_No response_

### Version

a24318ee

### Build Type

- [x] Release
- [ ] Debug

### Additional context

_No response_

ModMaker101:
## Description
Remove breeding cap for all animals by setting limits to 0.

### Previous Behavior
You would get the "This animal can't enter love mode..." message when the limits were reached.

### Root Cause
Hardcoded limits in `MobCategory.h`.

### New Behavior
No more limits.

### Fix Implementation
I set all the breeding and spawn egg limits to 0 in `MobCategory.h`.

### AI Use Disclosure
None.

## Related Issues
- Fixes #1416


codeHusky:
It'd be better if you expanded this out into a config (for both client and dedi server) rather than just nuking the hardcoded limits, especially for future console support. Won't merge this as-is, so changing to a draft.

ModMaker101:
If we're seeking parity with Java edition, then we would need to remove the breeding limits. Unless the purpose of a config would be to continue to support lower-end hardware/keep legacy "features."

PUFFMON:
> If we're seeking parity with Java edition, then we would need to remove the breeding limits.
>
Jave Edition already allows you to modify different game values through gameRules, so it wouldn't be too far from vanilla, Plus the config would be a good addition for servers to avoid lag and stuff

codeHusky:
> If we're seeking parity with Java edition

We're not looking for full JE parity. We're trying to maintain the base LCE experience by default, while also enabling configuring / expanding upon what's normally possible in LCE easily. Configuration of some kind is the best way to go.

ModMaker101:
Good point, but what are you looking for as a config? I personally don't think the scope of this PR is appropriate to implement a gamerule-like system. So with that, I get while it's a draft, but is there active development towards a system? And if/when that system gets implemented should I then add a property or similar for the mob count?

Jazzitch:
Maybe setting to 250 might be better because the project wants to keep console aspects in mind.

ModMaker101:
Maybe as a temporary fix, but I don't think that setting to a higher-but-finite cap would be the best solution. Now, I am open to different opinions, but like husky said it'd be better if we let kept it configurable. Also, for the console aspect... I don't think that we will ever be able to come back to consoles in the foreseeable future, but I haven't been on this project in a hot minute, so maybe that's a plan. But really, a lot has happened 🫪.

PR #1535, opened 2026-04-21
4JLibs as a submodule

Patoke:
## Description
Directly pulls from the 4JLibs repository on checkout as a submodule to allow for automated 4JLibs builds

### Previous Behavior
Precompiled libraries were used in Minecraft.Client/Windows64/lib

### Root Cause
The cmake design expected libraries to be then linked, similarly to how the original source did it

### New Behavior
Automatically builds and links 4JLibs independent of the platform, no extra files needed on the source tree

### AI Use Disclosure
No AI was used


Patoke:
Do note that actions do succeed with the ``nightly.yml`` and ``nightly-server.yml`` changes, submodules aren't being pulled by these actions which is why it fails in this PR, you can see they succeeded on my fork

PR #1542, opened 2026-04-24
Feat/custom recipes

DrPerkyLegit:
## Description

adding a packet for the server to sync recipes it has with the client

### Previous Behavior

before this the client and server just assumed they had the correct recipe list, now its synced to ensure proper crafting

### Root Cause

client and server were always the same, tight protocol versions ensured this was never needed on consoles
### New Behavior

now the client gets a list of recipes from the server
### Fix Implementation

during the PreLoginPacket the client sets a dirty recipe state thats waiting for the recipe packet, its expecting the packet before the final LoginPacket from the server, the server then sends a CustomPayloadPacket with encoded recipe data making the client no longer wait for recipes and continue, if the server doesnt send the recipe data before the final LoginPacket then it will rebuild recipes from the clients list of known recipes
### AI Use Disclosure

no
## Related Issues
- Fixes #[issue-number]
- Related to #[issue-number]


DrPerkyLegit:
i have tested this and it looks to be working perfectly but i would prefer someone else to test this out a bit before it gets accepted, we sync normal recipes so if you have issues with a default crafting recipe its caused by the syncing

PR #1543, opened 2026-04-25
Add configurable keyboard rebinding via controls.txt

### LINKED ISSUE #128: [ACCESSIBILITY] Keyboard/mouse rebinding.
Currently there's no way to rebind keyboard and mouse buttons.

### LINKED ISSUE #749: [Suggestion]  Add Customizable Keyboard Bindings
### Describe the solution you'd like

I tend to play minecraft with my shift key as sprint button, but when i got ur port of the legacy console edition, i realized i cldn't change bindings from the menu. I dived into the files for any bindings doc but cldn't find any. can you please add the bindings as a feature? It's annoying for certain people to use ctrl for sprint...thx for noticing. Cool Port!

### Is your suggestion related to a problem?

I am very frustrated when I use ctrl for sprinting...its just very annoying for me and some people out there.

### Additional context

_No response_

Lisiowen:
## Description
This PR adds a keyboard keybind configuration system via a `controls.txt` file in the game directory (similar to `username.txt`).

- Accessibility for users with **non-QWERTY** keyboard layouts (e.g France -> AZERTY)
- Support for **assistive input devices** for users with limited mobility
- QoL for users coming from Java/Bedrock who are used to their custom keybinds.
- Maintains default keybinds since the `controls.txt` file is **optional**, it only overrides the defaults rather than replacing the current implementation.

### Previous Behavior
Keyboard inputs were defined as **hardcoded constants** in the `KeyboardMouseInput.h` file, so there was no method to change these bindings without modifying the source code and recompiling.

### Root Cause
The input system used a static, **read-only** implementation. There was also no system in place to bridge an external config file to the **KeyboardMouseInput** variables.

### New Behavior
On launch, the game attempts to load the `controls.txt` file. Users can override a keybind using **standard characters** or modifier keys using decimal **Windows Virtual-Key** codes.

-  `KEY_INVENTORY=I` -> Standard character for "I"
- `KEY_SPRINT=160` -> VK code for `Left Shift`
- `KEY_SNEAK=162` -> VK code for `Left Control`

If the file doesn't exist or the specific action isn't included in the file, the game defaults to the **standard LCE keybinds**.

### Fix Implementation
- Replaced the **hardcoded const members** in `KeyboardMouseInput.h` with **mutable static variables** to allow the values to be reassigned at runtime during the `KeyboardMouseInput::Init()` sequence without changing their global state.
- Added an `INPUT_BINDING_TABLE` macro as the source of truth for all the configurable inputs to `KeyboardMouseInput.cpp`, this defines the **default values** as well as generating the **lookup map** for any user configured overrides.
- Added the `LoadUserControls` function, this uses `std::map` for efficient lookups and `isdigit()` to differentiate between **VK codes** and **standard character** inputs. It's also used to check if the `controls.txt` file exists and for skipping over any empty lines in the file.
- Added the function call at the end of `KeyboardMouseInput::Init()` to ensure controls are ready before running.

### Cross-Platform + Wine Support
- Tested on both **Windows** and **Arch Linux** (Proton)
- Changes maintain the use of **Windows Virtual-Key** codes which translate consistently across compatibility layers
- Added logic to handle **carriage returns** (**\r**) so `controls.txt` files created on **unix-based** systems (Linux/macOS) or **Windows** don't cause parsing errors

### AI Use Disclosure
No AI used.

## Related Issues
Fixes #128
Closes #749

## Documentation + Templates
Thanks for looking over my PR, I'm happy to update the project `README` with a section on how to use `controls.txt` and I can make a **template file** as a starting point for users as well as listing common **VK codes**.


minecraftlover4183:
Thank you all so much for spreading my suggestion! it really is annoying!

